PR #49, opened 2015-03-17
rssbee: add ability to skip the already existing rss entries

Profpatsch:
Helpful if you want to forward to irc for example (no spamming when bot starts up).

### DIFF
--- a/bees/rssbee/rssbee.go
+++ b/bees/rssbee/rssbee.go
@@ -33,14 +33,11 @@ import (
 type RSSBee struct {
 	bees.Bee
 
-	// options
-	url        string
-	skip_first bool
-
-	eventChan chan bees.Event
-
+	url string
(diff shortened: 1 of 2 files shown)


bees/rssbee/rssbee.go
```diff
@@ -24,19 +24,23 @@
...
 	"time"
 )
 
 type RSSBee struct {
 	bees.Bee
 
-	url         string
+	// options
+	url        string
+	skip_first bool
```
muesli: skip_frist isn't used anywhere iny our code. What is it for?

Profpatsch:
You right.

bees/rssbee/rssbee.go
Profpatsch: You’re right.

PR #50, opened 2015-04-17
Implement exec bee

dschmidt:


bees/execbee/execbee.go
```diff
@@ -0,0 +1,120 @@
...
+ *    GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *    Authors:
+ *      Dominik Schmidt <domme@tomahawk-player.org>
+ */
+
+// beehive's IRC module.
```
muesli: Really? ;-)

bees/execbee/execbee.go
```diff
@@ -0,0 +1,120 @@
...
+					errScanner := bufio.NewScanner(errReader)
+					go func() {
+						for errScanner.Scan() {
+							foo := errScanner.Text()
+							log.Println("execbee: err: | ", foo)
+							errBuffer = append(errBuffer, foo)
+						}
+					}()
+
+					cmd.Start()
```
muesli: Start may fail - e.g. if a command doesn't exist. Should check for error returned for Start() and Wait() below, too.

bees/execbee/execbee.go
```diff
@@ -0,0 +1,120 @@
...
+					cmd.Wait()
+
+					ev := bees.Event{
+						Bee:  mod.Name(),
+						Name: "cmd",
```
muesli: Should maybe be named differently. "commandResult" or something?

dschmidt: I called it that way because the struct is called ```cmd``` in go too

dschmidt: https://golang.org/pkg/os/exec/#Cmd

PR #61, opened 2016-12-06
Adds regexp filter

rubiojr:
Use regular expressions to filter in chains:

```
{
   "Filter":{
      "Name":"matches",
      "Options":[
         {
            "Name":"text",
            "Value":"^foo.*$"
         }
      ]
   }
}
```


muesli:
Love it! Just in case you guys like IRC, too, feel free to join me/us on freenode, channel #beehive.

PR #82, opened 2017-02-13
Making sense of the Time Cron UI

sks:
Not sure How to run this on local.

I tried ` go run beehive.go ` but no luck.

Is there a developer documentation on how to run beehive on local with the new codes?

### DIFF
--- a/bees/cronbee/cronbeefactory.go
+++ b/bees/cronbee/cronbeefactory.go
@@ -71,17 +71,17 @@ func (factory *CronBeeFactory) Options() []bees.BeeOptionDescriptor {
 	opts := []bees.BeeOptionDescriptor{
 		{
 			Name:        "second",
-			Description: "00-59 for a specific time; * for ignore",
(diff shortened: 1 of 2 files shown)


muesli:
@sks I assume the problem you're facing is, that you're editing files in your fork, but when you "go run" or "go build" the imports in beehive.go will still point to my repository - hence your changes won't get compiled.

bees/cronbee/cronbeefactory.go
```diff
@@ -91,12 +91,12 @@ func (factory *CronBeeFactory) Options() []bees.BeeOptionDescriptor {
 		},
 		{
 			Name:        "day_of_month",
-			Description: "01-31 for a specific time; * for ignore",
+			Description: "01-31 for a specific day of month; * for ignore",
```
muesli: I would call this "day of the month" (note the extra 'the') as you did in timebee, too.

sks: Made the change

Issue #70, opened 2017-02-11
Add url parameter to Twitter events


muesli:
```url``` should point to the canonical Twitter URL of the tweet.

seblw:
hi, could you explain more what needs to be done here?

muesli:
Right now the twitter events usually contain the tweet/event's author and the content of the tweet. For many use-cases it would be nice to also get the url of the tweet on twitter.com - so you can link to it.

All that really needs be done is retrieving the URL from the API and passing it along with the event, like the other params.

seblw:
Ok, thanks, I'll work on it.

muesli:
Nice! Should you have more questions, don't hesitate to post them here - or join us on Gitter/IRC.

seblw:
#84

PR #84, opened 2017-02-13
Add url parameter to Twitter events

seblw:
Example log output:

INFO[1566] Event received: Twittertest / tweet - is triggered whenever someone you follow tweets 
INFO[1566] 	Options: {username string mcclure111}       
INFO[1566] 	Options: {text string RT @tweegeemee: 170213_063132_b.clj https://t.co/6XDuJuYtWQ #ProceduralArt #generative https://t.co/7kgwYxDg8G} 
INFO[1566] 	Options: {url string https://twitter.com/statuses/831266508819030019}

### DIFF
--- a/bees/twitterbee/twitterbee.go
+++ b/bees/twitterbee/twitterbee.go
@@ -171,6 +171,11 @@ func (mod *TwitterBee) handleStreamEvent(item interface{}) {
 					Type:  "string",
 					Value: status.Text,
 				},
+				{
+					Name:  "url",
+					Type:  "url",
+					Value: "https://twitter.com/statuses/" + status.IdStr,
+				},
 			},
 		}
 
@@ -214,6 +219,11 @@ func (mod *TwitterBee) handleStreamEvent(item interface{}) {
 					Name:  "text",
 					Type:  "string",
 					Value: status.TargetObject.Text,
+				},
+				{
+					Name:  "url",
+					Type:  "url",
+					Value: "https://twitter.com/statuses/" + status.TargetObject.IdStr,
 				},
 			},
 		}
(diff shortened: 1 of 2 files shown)


seblw:
#70

bees/twitterbee/twitterbee.go
```diff
@@ -171,6 +171,11 @@ func (mod *TwitterBee) handleStreamEvent(item interface{}) {
 					Type:  "string",
 					Value: status.Text,
 				},
+				{
+					Name:  "url",
+					Type:  "string",
+					Value: "https://twitter.com/statuses/" + status.IdStr,
+				},
```
muesli: The same needs be done for the EventTweet case below.

bees/twitterbee/twitterbeefactory.go
```diff
@@ -168,6 +168,11 @@ func (factory *TwitterBeeFactory) Events() []bees.EventDescriptor {
 					Description: "text content of the tweet",
 					Type:        "string",
 				},
+				{
+					Name:        "url",
+					Description: "URL of the tweet",
+					Type:        "string",
```
muesli: Use the type "url" for those url parameters, please. I know it's not consistently used across all bees yet, but I'm about to make that switch in the near future.

seblw: ok, should the name be changed to something like twitt_url to avoid confusion or it's ok as is?

muesli: The name 'url' is totally fine. It's namespaced to this Bee, anyway, and eventually the type shouldn't be a user-facing value, so I don't mind it being a name and type at the same time.

PR #94, opened 2017-02-17
Added Twilio Bee

jamesbvaughan:
This Bee allows you to send SMS messages through Twilio's API.

It makes use of @carlosdp's Twilio Go helper library [twiliogo](https://github.com/carlosdp/twiliogo).


bees/twiliobee/twiliobee.go
```diff
@@ -0,0 +1,72 @@
...
+func (mod *TwilioBee) Action(action bees.Action) []bees.Placeholder {
+	outs := []bees.Placeholder{}
+
+	switch action.Name {
+	case "send":
+		body := ""
+
+		action.Options.Bind("body", &body)
+
+		twilio.NewMessage(mod.client, mod.from_number, mod.to_number, twilio.Body(body))
```
muesli: Idea: we could make ```to_number``` an option of this action, too. That would allow you to set different receivers for different events. (can be a separate pull request though)

jamesbvaughan: Ahh, I like that idea. I'll do it in a separate pull request. Do you think it should be an option for the bee and the action (so that you could set a default) or just for the action?

muesli: Would be nice to have both!

jamesbvaughan: Cool. If I have time later today I'll do that, but otherwise I'll do it next week.

muesli:
Nice work, thank you!

bees/twiliobee/twiliobee.go
PedroMD: I was just looking into this bee and my first thought was precisely (why isn't this part of the action ?!) 

Nice work so far - will wait for that PR :)

jamesbvaughan: Ahh thanks for the bump. I had almost forgotten about this!

PR #98, opened 2017-02-19
Added (un)follow for the tumblrbee

PenguWin:
Added logging on error and implemented follow/unfollow for the tumblrbee


adaawg:
Well done

PR #99, opened 2017-02-21
added notification wrapper for os x

akashshinde:


bees/notificationbee/notificationbee_osx.go
```diff
@@ -23,7 +23,11 @@
 // Package notificationbee is a Bee that can trigger desktop notifications.
 package notificationbee
 
+import "github.com/deckarep/gosx-notifier"
+
 // Run executes the Bee's event loop.
 func (mod *NotificationBee) execAction(text string, urgency uint32) {
-	//FIXME: implement
+	note := gosxnotifier.NewNotification(text)
+	note.Subtitle = "this is subtitle"
```
muesli: We should set something reasonable here.

akashshinde: ohh yeah, I think as title we can have as Beehive and subtitle as text

PR #101, opened 2017-02-22
Jenkins bee will trigger event when last build status gets changed

akashshinde:


muesli:
Can you explain what this change does / how it works?

muesli:
In particular what's the difference between status and build status for Jenkins?

akashshinde:
Basically It will trigger event if any one of the Jenkins job started it's build or ended build.

bees/jenkinsbee/jenkinsbee.go
```diff
@@ -28,8 +28,9 @@ import (
 	"io/ioutil"
 	"net/http"
 	"time"
-
+	//"github.com/bndr/gojenkins"
```
muesli: If we don't use it yet, remove this line or add a comment like "worth investigating and noteworthy for future updates".

bees/jenkinsbee/jenkinsbee.go
```diff
@@ -80,8 +86,38 @@ func (mod *JenkinsBee) announceStatusChange(j Job) {
...
+				Value: strconv.FormatBool(j.LastBuild.Building),
+			},
+		},
+	}
+	mod.eventChan <- event
+}
+
 func (mod *JenkinsBee) poll() {
-	request, err := http.NewRequest("GET", mod.url+"/api/json", nil)
+	request, err := http.NewRequest("GET", mod.url+"/api/json?tree=jobs[name,url,color,lastBuild[building]]", nil)
```
muesli: This is just filtering the output we want to retrieve, right? In other words: it will still report the data we need for the previously existing "status_change" event?

bees/jenkinsbee/jenkinsbee.go
```diff
@@ -153,6 +191,12 @@ func (mod *JenkinsBee) triggerBuild(jobname string) {
 	}
 }
 
+//func (mod *JenkinsBee ) pollLatestBuild() {
```
muesli: If it's not used yet, remove it or add a comment why this code may be useful or will be required in the near future.

bees/jenkinsbee/jenkinsbeefactory.go
```diff
@@ -114,6 +114,33 @@ func (factory *JenkinsBeeFactory) Events() []bees.EventDescriptor {
...
+					Description: "Current status of the job ('red' or 'blue')",
+					Type:        "string",
+				},
+				{
+					Name:        "url",
+					Description: "URL of the affected job",
+					Type:        "string",
+				},
+				{
+					Name:        "building",
```
muesli: Does this mean the value of ```building``` is either the string 'started' or the string 'finished'? Or is it actually a boolean value?

Also: note the upper-case 'F' in 'Finished' vs. the lower-case 'started'.

bees/jenkinsbee/jenkinsbee.go
akashshinde: yes, status_change event is working fine with it.

bees/jenkinsbee/jenkinsbee.go
akashshinde: removed

bees/jenkinsbee/jenkinsbeefactory.go
akashshinde: it returns true or false in string format only

muesli: If that's the case, I'd suggest converting it to a proper boolean here.

akashshinde: boolean will work ? I though it will work only for string type

akashshinde: I am getting it already in bool, so I don't think so we need to use `strconv` package here

PR #108, opened 2017-02-28
Switch to go-mail in EmailBee

henson:
use go-mail to send HTML mail


bees/emailbee/emailbee.go
```diff
@@ -43,18 +44,32 @@ func (mod *EmailBee) Action(action bees.Action) []bees.Placeholder {
...
+		action.Options.Bind("text", &mailtext)
 		action.Options.Bind("subject", &subject)
 
-		text = "Subject: " + subject + "\n\n" + text
-		auth := smtp.PlainAuth("", mod.username, mod.password, mod.server[:strings.Index(mod.server, ":")])
-		err := smtp.SendMail(mod.server, auth, mod.username, []string{to}, []byte(text))
-		if err != nil {
+		var host string
+		var port int
+		if strings.Index(mod.server, ":") != -1 {
```
muesli: Instead of doing this manually, I'd suggest using net.SplitHostPort: https://golang.org/pkg/net/#SplitHostPort

muesli:
Thanks, nice work! I think we should add two action parameters in that case, though: "text" for the plaintext content of the mail, and "html" for the HTML content. Will merge after that change, so we don't break existing behavior & functionality.

bees/emailbee/emailbee.go
henson: OK

henson:
update emailbee add an option for type choice

muesli:
Sorry, that's not quite what I meant. You usually want to embed both, a plaintext version and an HTML version, when sending an email. HTML-only is rather weird. So instead of adding a content-type option, add an "html" option, which (if available) gets set as the text/html content of the mail.

muesli:
Also, just a heads up: please always run 'go fmt' on your code before committing.

henson:
I hope the last commit is that you want. And I also go fmt the code before committing

bees/emailbee/emailbeefactory.go
```diff
@@ -113,13 +113,20 @@ func (factory *EmailBeeFactory) Actions() []bees.ActionDescriptor {
...
+					Mandatory:   true,
 				},
 				{
 					Name:        "text",
 					Description: "Content of the email",
 					Type:        "string",
 					Mandatory:   true,
 				},
+				{
+					Name:        "HTML",
```
muesli: Make that a lower-case "html" like all other options, please.

bees/emailbee/emailbeefactory.go
```diff
@@ -113,13 +113,20 @@ func (factory *EmailBeeFactory) Actions() []bees.ActionDescriptor {
...
+					Description: "Whether to send email with HTML",
+					Type:        "bool",
```
muesli: This should be a "string" type.

bees/emailbee/emailbee.go
```diff
@@ -43,18 +45,40 @@ func (mod *EmailBee) Action(action bees.Action) []bees.Placeholder {
...
+			host = strings.Split(mod.server, ":")[0]
+			port, _ = strconv.Atoi(strings.Split(mod.server, ":")[1])
+		} else {
+			host = mod.server
+			port = 587
+		}
+
+		re, _ := regexp.Compile("\\<[\\S\\s]+?\\>")
```
muesli: Don't try to manually strip the tags. Use "text" as the plaintext content and "html" as the text/html content of the mail.

bees/emailbee/emailbeefactory.go
```diff
@@ -116,9 +116,13 @@ func (factory *EmailBeeFactory) Actions() []bees.ActionDescriptor {
...
 					Name:        "text",
-					Description: "Content of the email",
+					Description: "Content of the email using plain text",
+					Type:        "string",
+				},
+				{
+					Name:        "html",
+					Description: "Content of the email using HTML",
 					Type:        "string",
-					Mandatory:   true,
```
muesli: Make the "text" option mandatory and the html not-mandatory. Each email should should have a plaintext version and optionally an HTML version (as per standards).

henson:
update again, pls check

bees/emailbee/emailbeefactory.go
henson: OK

muesli: Actually, we could even make neither Mandatory and allow empty emails with just a subject. I've seen people do that (for better or worse).

muesli: Actually, we could even make neither Mandatory and allow empty emails with just a subject. I've seen people do that (for better or worse).

PR #114, opened 2017-03-08
fix something

henson:
1.emailbee.go: to remove the HTML tags for plainText
2.alertoverbee.go: remove, not replace
**3.chains_delete.go: I noticed that the chain is deleted but the action still remained, so I do the change is to delete chain's actions while deleting the chain **

### DIFF
--- a/bees/emailbee/emailbee.go
+++ b/bees/emailbee/emailbee.go
@@ -23,7 +23,7 @@ package emailbee
 
 import (
 	"net"
-
+	"regexp"
 	"strconv"
 
 	"github.com/muesli/beehive/bees"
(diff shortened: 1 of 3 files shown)


bees/emailbee/emailbee.go
```diff
@@ -55,9 +55,12 @@ func (mod *EmailBee) Action(action bees.Action) []bees.Placeholder {
 		m.SetHeader("From", mod.username)
 		m.SetHeader("To", to)
 		m.SetHeader("Subject", subject)
-		if plainText != "" {
-			m.SetBody("text/plain", plainText)
-		}
+
+		// remove the HTML tags, such as <html></html> and so on
+		re, _ := regexp.Compile("\\<[\\S\\s]+?\\>")
```
muesli: Why would that be needed? Whatever the user passes into "plaintext", we should use it and don't change the contents.

muesli:
Thanks for the pull request, but please don't combine multiple changes into one pull request. In this case I'm fine with the AlertOverBee fix, but I have a slightly different plan for the delete-chain fix - and now I can't merge the fix I would want.

henson:
OK, I change the pull request

PR #115, opened 2017-03-08
Added join/part/quit events to IRCBee

momo-aux:
Status change of a user can now be used as an event

### DIFF
--- a/bees/ircbee/ircbee.go
+++ b/bees/ircbee/ircbee.go
@@ -125,6 +125,42 @@ func (mod *IrcBee) part(channel string) {
 	}
 }
 
+func (mod *IrcBee) statusChange(eventChan chan bees.Event,conn *irc.Conn, line *irc.Line) {
+	//Line.CMD eq Handler Name ex: JOIN
+	message := ""
+	switch line.Cmd {
+	case "JOIN":
+		message = "User joined to channel " + line.Args[0]
+	case "PART":
+		message = "User parted to channel " + line.Args[0]
+	case "QUIT":
+		message = "User quit from channel " + line.Args[0]
+	default:
+		mod.LogErrorf("Unknown command " + line.Cmd + " in statusChange")
+		return
+	}
+	mod.Logln(message)
+	channel := line.Args[0]
+	user := line.Src[:strings.Index(line.Src, "!")]
+	ev := bees.Event{
+		Bee:  mod.Name(),
+		Name: strings.ToLower(line.Cmd),
+		Options: []bees.Placeholder{
+			{
+				Name:  "channel",
+				Type:  "string",
+				Value: channel,
+			},
+			{
+				Name:  "user",
+				Type:  "string",
+				Value: user,
+			},
+		},
+	}
+	eventChan <- ev
+}
+
 // Run executes the Bee's event loop.
 func (mod *IrcBee) Run(eventChan chan bees.Event) {
 	if len(mod.server) == 0 {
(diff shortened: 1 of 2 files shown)


bees/ircbee/ircbee.go
```diff
@@ -156,6 +156,55 @@ func (mod *IrcBee) Run(eventChan chan bees.Event) {
...
 	})
+
+	statusChange := func (conn *irc.Conn, line *irc.Line) {
+		//Line.CMD eq Handler Name ex: JOIN
+		message := ""
+		switch line.Cmd {
+		case "JOIN":
+				message = "User Joined to channel " + line.Args[0]
+		case "PART":
+				message = "User Joined to channel " + line.Args[0]
```
muesli: This should probably read "User parted".

bees/ircbee/ircbee.go
```diff
@@ -156,6 +156,55 @@ func (mod *IrcBee) Run(eventChan chan bees.Event) {
 	mod.client.HandleFunc("disconnected", func(conn *irc.Conn, line *irc.Line) {
 		mod.connectedState <- false
 	})
+
+	statusChange := func (conn *irc.Conn, line *irc.Line) {
```
muesli: I would declare this method outside of Run(). Makes the code easier to read and the inlining provides no advantages here.

bees/ircbee/ircbee.go
```diff
@@ -156,6 +156,55 @@ func (mod *IrcBee) Run(eventChan chan bees.Event) {
...
+		//Line.CMD eq Handler Name ex: JOIN
+		message := ""
+		switch line.Cmd {
+		case "JOIN":
+				message = "User Joined to channel " + line.Args[0]
+		case "PART":
+				message = "User Joined to channel " + line.Args[0]
+		case "QUIT":
+				message = "User Quit from channel " + line.Args[0]
+		}
```
muesli: What if ```line.Cmd``` is none of the above? A default case should probably abort/return here.

bees/ircbee/ircbee.go
```diff
@@ -156,6 +156,55 @@ func (mod *IrcBee) Run(eventChan chan bees.Event) {
...
+		mod.Logln(message)
+		channel := line.Args[0]
+		if channel == mod.client.Config().Me.Nick {
```
muesli: Since we're only handling channel related events and you can't join/part a query, this can probably be removed.

bees/ircbee/ircbeefactory.go
```diff
@@ -131,6 +131,57 @@ func (factory *IrcBeeFactory) Events() []bees.EventDescriptor {
...
 			},
 		},
+		{
+				Namespace:   factory.Name(),
+				Name:        "join",
+				Description: "A user joined a channel",
+				Options: []bees.PlaceholderDescriptor{
+					{
+						Name:        "user",
+						Description: "The username to watch",
```
muesli: Should read "The username that joined a channel".

bees/ircbee/ircbeefactory.go
```diff
@@ -131,6 +131,57 @@ func (factory *IrcBeeFactory) Events() []bees.EventDescriptor {
...
+						Type:        "string",
+					},
+					{
+						Name:        "channel",
+						Description: "The channel the user should be watched",
```
muesli: Should read "The channel the user joined".

bees/ircbee/ircbeefactory.go
```diff
@@ -131,6 +131,57 @@ func (factory *IrcBeeFactory) Events() []bees.EventDescriptor {
...
+						Type:        "string",
+					},
+				},
+			},
+			{
+					Namespace:   factory.Name(),
```
muesli: The indentation here got messed up. Always run ```go fmt *.go``` on your code before committing, to avoid such issues.

bees/ircbee/ircbee.go
```diff
@@ -125,6 +125,42 @@ func (mod *IrcBee) part(channel string) {
...
+	message := ""
+	switch line.Cmd {
+	case "JOIN":
+		message = "User joined to channel " + line.Args[0]
+	case "PART":
+		message = "User parted to channel " + line.Args[0]
+	case "QUIT":
+		message = "User quit from channel " + line.Args[0]
+	default:
+		panic("Unknown command " + line.Cmd + " in statusChange")
```
muesli: I don't think we should panic() here. It's just a message we don't handle / we're not interested in (yet). You can print out an error or even silently ignore it.

muesli:
Thank you, really appreciated!

PR #116, opened 2017-03-10
MumbleBee, a bee that connects to mumble servers

momo-aux:
… to channels

on a mumble Server.


bees/mumblebee/mumblebee.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+	//*/
+
+	config.Attach(gumbleutil.AutoBitrate)
+
+	connecting := false
+	disconnected := true
+	waitForDisconnect := false
+	for {
+		// loop on IRC connection events
```
muesli: Heh, classic copy/paste error.

bees/mumblebee/mumblebee.go
```diff
@@ -0,0 +1,250 @@
...
+		action.Options.Bind("user", &user)
+		action.Options.Bind("text", &text)
+
+		mod.client.Users.Find(user).Send(text)
+
+	case "send_to_channel":
+
+		channel, text := "", ""
+
+		action.Options.Bind("channel", &channel)
```
muesli: ```channel``` doesn't seem to be used here. What's the difference to the "say" action below?

bees/mumblebee/mumblebee.go
```diff
@@ -0,0 +1,250 @@
...
+				mod.Logln("Connected to:", mod.server)
+				if err != nil {
+					mod.Logln("Failed to connect to Mumble:", mod.server, err)
+					connecting = false
+				}
+			}
+		}
+		select {
+		case status := <-mod.connectedState:
+			mod.Logln("status case")
```
muesli: Probably best to remove this and the other ```status xxx``` debug/development log-messages below so we don't spam our user's logs.

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,203 @@
...
+func (factory *MumbleBeeFactory) Options() []bees.BeeOptionDescriptor {
+	opts := []bees.BeeOptionDescriptor{
+		{
+			Name:        "address",
+			Description: "Hostname of Mumble server, eg: de.clanwarz.com:64738",
+			Type:        "string",
+			Mandatory:   true,
+		},
+		{
+			Name:        "channel",
```
muesli: Can you join multiple channels concurrently on a Mumble server? In this case, we could make it ```channels``` of ```[]string``` type. We could also get rid of the ```move``` action then.

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,203 @@
...
+// Events describes the available events provided by this Bee.
+func (factory *MumbleBeeFactory) Events() []bees.EventDescriptor {
+	events := []bees.EventDescriptor{
+		{
+			Namespace:   factory.Name(),
+			Name:        "message",
+			Description: "A message was received over Mumble",
+			Options: []bees.PlaceholderDescriptor{
+				{
+					Name:        "text",
```
muesli: We could also emit the channel this message was received on.

hives.go
```diff
@@ -49,4 +49,5 @@ import (
 	_ "github.com/muesli/beehive/bees/twiliobee"
 	_ "github.com/muesli/beehive/bees/twitterbee"
 	_ "github.com/muesli/beehive/bees/webbee"
+	_ "github.com/muesli/beehive/bees/mumblebee"
```
muesli: Sort them alphabetically. I suggest running ```go fmt *.go``` which does that for you.

bees/mumblebee/mumblebee.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	config.Attach(gumbleutil.Listener{
+		TextMessage: func(e *gumble.TextMessageEvent) {
+			ev := bees.Event{
+				Bee:  mod.Name(),
+				Name: "message",
+				Options: []bees.Placeholder{
+					{
+						Name:  "channels",
```
muesli: This option isn't defined in the factory. I also assume it should be ```channel``` (singular) of type ```string``` in that case.

bees/mumblebee/mumblebee.go
```diff
@@ -140,7 +135,7 @@ func (mod *MumbleBee) Run(eventChan chan bees.Event) {
 				Name: "message",
 				Options: []bees.Placeholder{
 					{
-						Name:  "channels",
+						Name:  "channel",
 						Type:  "[]string",
```
muesli: Should be ```string``` then, not an array of strings.

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,203 @@
...
+	}
+	return events
+}
+
+// Actions describes the available actions provided by this Bee.
+func (factory *MumbleBeeFactory) Actions() []bees.ActionDescriptor {
+	actions := []bees.ActionDescriptor{
+		{
+			Namespace:   factory.Name(),
+			Name:        "say",
```
muesli: This Action can be removed as it doesn't exist any longer.

bees/mumblebee/mumblebeefactory.go
muesli: The ```channel``` is now being emitted, but is still missing here.

bees/mumblebee/mumblebee.go
```diff
@@ -0,0 +1,240 @@
...
+	config.Attach(gumbleutil.Listener{
+		TextMessage: func(e *gumble.TextMessageEvent) {
+			ev := bees.Event{
+				Bee:  mod.Name(),
+				Name: "message",
+				Options: []bees.Placeholder{
+					{
+						Name:  "channel",
+						Type:  "[]string",
+						Value: e.Channels,
```
muesli: I'm confused. ```e.Channels``` seems to indicate a message can be received on multiple channels? Isn't it always just one channel?

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,208 @@
...
+				},
+				{
+					Name:        "user",
+					Description: "The user that sent the message",
+					Type:        "string",
+				},
+				{
+					Name:        "channels",
+					Description: "The channels where the message was received",
+					Type:        "string",
```
muesli: This sould be a ```[]string```.

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,208 @@
...
+		{
+			Namespace:   factory.Name(),
+			Name:        "send_to_channel",
+			Description: "Sends a message to a channel",
+			Options: []bees.PlaceholderDescriptor{
+				{
+					Name:        "channel",
+					Description: "Which channel to send the message to",
+					Type:        "string",
+					Mandatory:   true,
```
muesli: This should be false. If no channel gets provided, this bee will send to the current channel.

bees/mumblebee/mumblebeefactory.go
```diff
@@ -0,0 +1,208 @@
...
+					Name:        "channel",
+					Description: "The channel to move in",
+					Type:        "string",
+					Mandatory:   true,
+				},
+			},
+		},
+		{
+			Namespace:   factory.Name(),
+			Name:        "play",
```
muesli: This action doesn't actually exist yet.

muesli:
Thank you @momo-aux!

Joshfindit:
P.s. Excellent move from the previous title to MumbleBee. Much clearer. (plus, I think this is a great name in general)

muesli:
@Joshfindit I'm not quite sure I understand what you're referring to?

Joshfindit:
@muesli:

`Bee that can react on text messages, send text messages and move…` = Confusing title. Sounds like SMS.
`MumbleBee, a bee that connects to mumble servers` = Much clearer.
`MumbleBee` = Catchy name.

muesli:
Understood & agreed now :)

PR #124, opened 2017-03-21
added cricket bee

akashshinde:


muesli:
As a first note: this does break compatibility with Go <1.5: https://travis-ci.org/muesli/beehive/builds/213439848

I'm not sure we really still need to support 1.3 and 1.4, but it seems it's only breaking because of strings.Compare(), which wasn't around before 1.5. This could easily be worked around, though. Any objections to changing that?

bees/cricketbee/cricketbee.go
```diff
@@ -0,0 +1,101 @@
...
+			Value: r.Overs,
+		},
+	}
+}
+
+func (c *CricketBee) announceCricketEvent(response gocricket.ResponseEvent) {
+	switch response.EventType {
+	case gocricket.EVENT_OUT:
+		c.beeEvt <- bees.Event{
+			Name:    "event_out",
```
muesli: Maybe we could drop the "event_" prefix from those names. It's an event by definition already.

bees/cricketbee/cricketbee.go
```diff
@@ -0,0 +1,101 @@
...
+	evt := make(chan gocricket.ResponseEvent)
+	// Start Cricket GoRoutine to poll cricket score
+	cricket := gocricket.NewCricketWatcher(c.favTeam, evt)
+	cricket.Start()
+	c.beeEvt = cin
+	for {
+		select {
+		case e := <-evt:
+			c.announceCricketEvent(e)
+		}
```
muesli: You need to add a case for ```mod.SigChan```. If you receive anything from that channel, stop running and return from ```Run()```. That way bees can get stopped gracefully.

hives.go
```diff
@@ -50,4 +50,5 @@ import (
 	_ "github.com/muesli/beehive/bees/twiliobee"
 	_ "github.com/muesli/beehive/bees/twitterbee"
 	_ "github.com/muesli/beehive/bees/webbee"
+	_ "github.com/muesli/beehive/bees/cricketbee"
```
muesli: Those imports are supposed to be sorted alphabetically. Simply run go fmt on all your code - it will take care of it for you!

muesli:
Nice work! Just minor nitpicks (as usual, heh).

hives.go
```diff
@@ -23,6 +23,7 @@ package main
 import (
 	_ "github.com/muesli/beehive/bees/alertoverbee"
 	_ "github.com/muesli/beehive/bees/anelpowerctrlbee"
+	_ "github.com/muesli/beehive/bees/cricketbee"
```
akashshinde: This path should not be taken from `muesli/beehive/bees/cricketbee` right ?

muesli: It should, absolutely, that's fine.

akashshinde: ah get it, I was thinking this cricket lib which I wrote, my bad

PR #127, opened 2017-03-29
Added Gitterbee

penguwin:
Gitterbee interacts with gitter.
Unfortunately only actions are supported by now
because I'm having some issues in the events part.
All event code has been commented out, maybe some of
you guy's will have a look at it.


muesli:
Thank you!

PR #134, opened 2017-04-08
[WiP] Augment installation/running instructions

rubiojr:
A bit torn on this, but added a couple of examples.
Perhaps too much content for the main README, Linux centric and not detailed enough?

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Connecting those modules with each other lets you create immensly useful agents.
 
 ![beehive's Logo](/assets/logo_256.png?raw=true)
 
-## Installation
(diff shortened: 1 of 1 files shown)


muesli:
Some outdated, but still a bunch of really useful information in here. Should probably move that to a wiki entry and link to it from the README.

rubiojr:
This is woefully outdated now, and not very useful probably.

rubiojr:
> Some outdated, but still a bunch of really useful information in here. Should probably move that to a wiki entry and link to it from the README

👌 Let me have a closer look.

PR #135, opened 2017-04-09
Add support for embedded assets

rubiojr:
`make embed` creates a beehive binary with embedded assets (js, img, css, etc).

A go-bindata stub (api/bindata.go) was added so embedding assets isn't required,
which could be useful in low memory devices.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,24 @@
+all: embed
+
+submodule:
+	git submodule update --init
+
+noembed: submodule build
+
+generate:
+	go-bindata --pkg api -o api/bindata.go --ignore \\.git assets/... config/...
+
+go-bindata:
+	[ -f $(shell go env GOPATH)/bin/go-bindata ] || go get -u github.com/jteeuwen/go-bindata/go-bindata
+
+embed: submodule go-bindata generate build
+
+build:
+	go build
+
+get-deps:
+	go get -u
+
+clean:
+	rm -f beehive
+.PHONY: clean embed go-bindata get-deps noembed generate submodule build all
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ beehive.conf
 
 # Jetbrains IDEs
 .idea
+
+# Embedded assets
+api/bindata.go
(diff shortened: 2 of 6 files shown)


rubiojr:
@muesli the RSS on my rpi with embedded assets is ~18MiB.

rubiojr:
Ready for some :eyes:.

`make` currently defaults to building without embedded assets, but we could embed by default to make deployments easier (i.e. because we don't need an additional step to manually copy assets and images).

muesli:
Looks like this broke the Travis-CI tests, as it's now just executing "make" and doesn't do the "go get" automatically. Will have to read up what's best practice here.

muesli:
I think having a "get-deps" Makefile target as described here makes the most sense: https://docs.travis-ci.com/user/languages/go/#Dependency-Management

Otherwise every time you build it would first try to update all dependencies - which might be unexpected.

rubiojr:
> Looks like this broke the Travis-CI tests, as it's now just executing "make" and doesn't do the "go get" automatically. Will have to read up what's best practice here.

@muesli 3a0d75be28823269c3096a6b276986694326902e will hopefully do the trick, according to https://docs.travis-ci.com/user/languages/go

muesli:
Looking good, will try it out!

Makefile
```diff
@@ -0,0 +1,18 @@
+all: noembed
+
+submodule:
+	git submodule update --init
+
+noembed: submodule build
+
+generate:
+	go get github.com/jteeuwen/go-bindata
```
muesli: This doesn't seem to build go-bindata. It seems to require a manual build with "make".

muesli: Oh, I think we need to get github.com/jteeuwen/go-bindata/go-bindata (note the extra go-bindata)

muesli:
We should also update the README.md build instructions & wiki. Related: https://github.com/muesli/beehive/pull/134

rubiojr:
@muesli I think I addressed the feedback. Lemme know what you think.

rubiojr:
> We should also update the README.md build instructions & wiki. Related: #134

README updated here. I'll update the wiki (https://github.com/muesli/beehive/wiki/Installation) once we merge this.

rubiojr:
Wiki page updated.

PR #142, opened 2017-04-14
Script to build "native" apps

rubiojr:
Use [nativefier](https://github.com/jiahaog/nativefier) to wrap the beehive binary and build a desktop app.


rubiojr:
We can build Linux apps with this too, but I haven't tried. The script will need to be modified to support it.

screen shot 2017-04-14 at 20 12 19

rubiojr:
@muesli we could start publishing binary assets with the releases at some point. At least for MacOS and Linux users, so they can easily run Beehive.

rubiojr:
Going to use `launchd` on macOS, creating a launch agent so we can easily monitor `beehived`, automatically launch it when the compute starts, easily kill it without the hacky killall, etc.

muesli:
Note: I think we can (and should) dynamically generate the beehive.icns

rubiojr:
> Note: I think we can (and should) dynamically generate the beehive.icns

:+1: I'll try to figure it out.

rubiojr:
> 👍 I'll try to figure it out.

Found something from Henry on stackoverflow: http://stackoverflow.com/a/20703594

rubiojr:
Accidentally closed via force-push 😢.

rubiojr:
I've resumed work on this and added a [snap](https://snapcraft.io) config so we can generate snap (via `script/build-app`) for the distributions that support it.
Still a WiP but functional, added in https://github.com/muesli/beehive/pull/142/commits/aba9788b58f9f893d5f439fa2667778caacf7dac

@muesli: would you be interested in registering as a developer on https://snapcraft.io so we can eventually publish the snap there?

rubiojr:

PR #145, opened 2017-04-19
Strip binary by default

rubiojr:
Reduces the binary size considerably (~15 MiB vs ~23 MiB):

```
[9204][rubiojr@octopro] make debug

[9205][rubiojr@octopro] ls -la beehive
-rwxr-xr-x  1 rubiojr  staff  23024756 Apr 19 23:10 beehive

[9205][rubiojr@octopro] make

[9205][rubiojr@octopro] ls -la beehive
-rwxr-xr-x  1 rubiojr  staff  15413708 Apr 19 23:11 beehive
```

Should we do this by default?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,9 @@ go-bindata:
 embed: submodule go-bindata generate build
 
 build:
+	go build --ldflags '-s -w'
+
+debug: submodule go-bindata generate
 	go build
 
 get-deps:


muesli:
I'm not quite sure why the Travis tests fail for Golang <1.6, but by the looks it might only be a problem since it's checking code in a git branch. Let's merge and see how it behaves when testing the code in git master.

muesli:
Yeah, exactly what I thought, worked fine once it's landed in master. Thanks for another nice PR!

PR #147, opened 2017-04-22
PoC: Updated Placeholder struct in devrantbee.go

penguwin:
This is more like a proof on concept.
'Generating' the []bees.Placeholder struct with the 'reflect' package.
Unfortunately I'm violating the beehive style standard with changing
some PlaceholderDescriptor names like 'ID' to 'Id'.
The PoC shows that we don't have to write all the Placeholder structs
in beehive per hand

### DIFF
--- a/bees/devrantbee/devrantbee.go
+++ b/bees/devrantbee/devrantbee.go
@@ -25,6 +25,8 @@ package devrantbee
 import (
 	"github.com/jayeshsolanki93/devgorant"
 
+	"reflect"
+
 	"github.com/muesli/beehive/bees"
 )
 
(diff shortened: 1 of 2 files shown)


bees/devrantbee/devrantbee.go
```diff
@@ -81,61 +83,30 @@ func (mod *DevrantBee) Action(action bees.Action) []bees.Placeholder {
...
+
+	for i := 0; i < v.NumField(); i++ {
+		names[i] = v.Type().Field(i).Name
+		types[i] = v.Field(i).Type().String()
+		values[i] = v.Field(i).Interface()
+
+		// Parsing the values into a bees.Placeholder struct
+		opts[i] = bees.Placeholder{
+			Name:  names[i],
+			Type:  types[i],
```
muesli: This could be a bit problematic, as we have some internal types that don't map to a Golang-type (e.g. `address` comes to mind) and vice versa. We could probably tag those fields in the structs, though.

bees/devrantbee/devrantbeefactory.go
```diff
@@ -74,52 +74,57 @@ func (factory *DevrantBeeFactory) Events() []bees.EventDescriptor {
 			Description: "is triggered after rants were fetched",
 			Options: []bees.PlaceholderDescriptor{
 				{
-					Name:        "ID",
+					Name:        "Id",
```
muesli: Looks like you renamed those by accident?

penguwin: Renaming here _was_ necessary because we're parsing the Placeholders name in devrantbee.go.
The returned names are the same as in the particular struct fields (in this case: [devrant](https://godoc.org/github.com/jayeshsolanki93/devgorant#RantModel)).

A simple solution should be to write/find a go library which parses 'camelCase' strings into 'snake_cases'. Therefore we could convert the structs field name just before we translate them as bees.Placeholder name and stay at json styled 'snake_cases' for our bees.Placeholder and bees.PlaceholderDescriptor.

bees/devrantbee/devrantbee.go
penguwin: I already knew this problem exists (as the title says it's just a proof of concept ;p) but I'll work on a solution to handle nested structs, arrays et cetera in a pleasant way.

PR #150, opened 2017-05-06
filesystem notification bee

rubiojr:
A bee that reacts to filesystem events and forwards them to other
bees interested.


rubiojr:
Wondering now if I should switch to https://github.com/rjeczalik/notify that supports recursivity and FSEvents on macOS (more efficient apparently).

rubiojr:
> Wondering now if I should switch to https://github.com/rjeczalik/notify that supports recursivity and FSEvents on macOS (more efficient apparently).

We talked about this on IRC. Recursivity (without proper control) is dangerous and kqueue is probably good enough for macOS and this particular case, so sticking to https://github.com/fsnotify/fsnotify for now.

Unrelated to the previous comment: it seems that the new `fsnotify` dep pulls `x/sys/unix` which [breaks go 1.3](https://github.com/golang/go/issues/9211). @muesli, how do you feel about building with go >= 1.4 and 🔥 1.3?

muesli:
That's a good question indeed. I guess after the 0.3 release (which should be imminent), I'd like to bump our Golang dependency to 1.7, so we can rely on the context package. Maybe we should keep the 1.3 dep for our 0.3 release and merge the fsnotify-bee afterwards?

rubiojr:
> Maybe we should keep the 1.3 dep for our 0.3 release and merge the fsnotify-bee afterwards?

👍

muesli:
I guess I need to re-enable 'go get'-ing all the dependencies (as opposed to relying on vendoring only), otherwise tests for such pull requests will fail. Sorry for that, I'll fix that on my side and merge your PR if you're ready.

rubiojr:
:bow:

PR #151, opened 2017-05-13
Adds S3 hive

rubiojr:
S3 bees can upload files to S3 compatible storage.


rubiojr:
This is a nice bee to combine with the [filesystem notification bee](https://github.com/muesli/beehive/pull/150)

rubiojr:
@muesli should I open a PR against https://github.com/muesli/beehive-vendor manually adding the new deps there or do we have tooling for that?

muesli:
I certainly wouldn't mind you opening a PR there, too. Thanks in advance! I currently use the `dep` tool for it. We should probably additionally fetch missing dependencies in the CI. This has the downside, that we won't get notified when deps are missing in `beehive-vendor` - but that's really only something we must not forget to check before releasing.

rubiojr:
> We should probably additionally fetch missing dependencies in the CI. This has the downside, that we won't get notified when deps are missing in beehive-vendor.

Either way works for me. CI failing is also a good reminder that we need to do something about `beehive-vendor`. I would have missed that repo was a thing otherwise.

muesli:
One nitpick: we must use square images / logos currently, otherwise they may get resized weirdly in the admin UI.

bees/s3bee/s3bee.go
```diff
@@ -0,0 +1,101 @@
...
+		objectPath := ""
+		action.Options.Bind("object_path", &objectPath)
+
+		if objectPath == "" {
+			objectPath = filepath.Base(path)
+		}
+
+		_, err := bee.client.FPutObject(bucket, objectPath, path, mime.TypeByExtension(filepath.Ext(path)))
+		if err != nil {
+			log.Fatalln(err)
```
muesli: Please use the logging methods provided by Beehive here. In this case: mod.LogFatal().

rubiojr:
Thanks @muesli, addressed feedback.

PR #153, opened 2017-05-13
Add timestamp to telegram events

rubiojr:
Minor but useful thingie so we can use timestamps with action templates.

### DIFF
--- a/bees/telegrambee/telegrambee.go
+++ b/bees/telegrambee/telegrambee.go
@@ -28,6 +28,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
 	telegram "github.com/go-telegram-bot-api/telegram-bot-api"
 
@@ -115,6 +116,11 @@ func (mod *TelegramBee) Run(eventChan chan bees.Event) {
 						Name:  "user_id",
 						Type:  "string",
 						Value: strconv.Itoa(update.Message.From.ID),
+					},
+					{
+						Name:  "timestamp",
+						Type:  "string",
+						Value: strconv.FormatInt(time.Now().Unix(), 10),
 					},
 				},
 			}
--- a/bees/telegrambee/telegrambee.go
+++ b/bees/telegrambee/telegrambee.go
@@ -119,8 +119,8 @@ func (mod *TelegramBee) Run(eventChan chan bees.Event) {
 					},
 					{
 						Name:  "timestamp",
-						Type:  "string",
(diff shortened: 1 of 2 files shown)


bees/telegrambee/telegrambee.go
```diff
@@ -116,6 +117,11 @@ func (mod *TelegramBee) Run(eventChan chan bees.Event) {
 						Type:  "string",
 						Value: strconv.Itoa(update.Message.From.ID),
 					},
+					{
+						Name:  "timestamp",
+						Type:  "string",
```
muesli: If we use unix timestamps here, we should treat the parameter as an integer as well. That being said, I think we actually wanna support handling timestamp as time.Time internally and let the conversion routine in placeholders.go do its magic (essentially make that support time.Time and let it convert from/to strings and integers).

rubiojr: Good idea.

muesli:
Now that you added a time format to Beehive, we could take another look at this!

rubiojr:
@muesli updated!

bees/telegrambee/telegrambee.go
```diff
@@ -116,6 +117,11 @@ func (mod *TelegramBee) Run(eventChan chan bees.Event) {
 						Type:  "string",
 						Value: strconv.Itoa(update.Message.From.ID),
 					},
+					{
+						Name:  "timestamp",
+						Type:  "time.Time",
```
muesli: I'd keep this simple and still call the type `timestamp`. Its internal Go type is separate to that type definition here, which we mostly use for the admin interface.

bees/telegrambee/telegrambeefactory.go
```diff
@@ -99,6 +99,11 @@ func (factory *TelegramBeeFactory) Events() []bees.EventDescriptor {
 					Description: "User ID sending the message",
 					Type:        "string",
 				},
+				{
+					Name:        "timestamp",
+					Description: "Timestamp",
+					Type:        "time.Time",
```
muesli: I'd keep this simple and still call the type `timestamp`. Its internal Go type is separate to that type definition here, which we mostly use for the admin interface.

rubiojr: Ahh gotcha, wasn't sure of that bit.

bees/telegrambee/telegrambee.go
rubiojr: Done.

PR #155, opened 2017-06-16
renames event in fsnotifybee

barbourkd:
When creating a chain triggered by an fsnotify event I was having an issue where none of the chains were being triggered. Dug into it some and found that in execChains the name of the triggered event and the name of the chain's event were not matching up. Looks like it was a mismatch between the event name in fsnotifybee and fsnotifybeefactory. Changing the names to match fixed this.


muesli:
Wow, thanks for noticing this!

PR #157, opened 2017-06-28
Updated cricketbee logo

penguwin:
Updated the logo .png to avoid having white borders around the logo


muesli:
Is this a royalty free picture we can use?

penguwin:
Well, I can't find an exact answer to this question, and I didn't checked that in the beginning.
So before we do any mistakes I'll just revoke the commit and replace the picture with another one which is marked as reuseable.
Is that ok? :)

PR #163, opened 2017-07-03
Ryanairbee

penguwin:
Opening pull request so far, i'll keep an eye of it.


muesli:
That means you don't consider it ready yet?

penguwin:
Well one the one hand, the ryanair api library does only offer two more functions at the moment: [GetAirportsCityCoordinatesMap](https://godoc.org/github.com/seblw/goryan#RyanairAPI.GetAirportsCityCoordinatesMap) and [GetAirportsCityIataMap](https://godoc.org/github.com/seblw/goryan#RyanairAPI.GetAirportsCityIataMap) and to be honest I don't think those are 'really' useful. I have no problem to implement them into the hive before we merge the pull request if desired.

On the other hand we should consider merging the major/bigger airlines into one 'flights/airlines' hive in further future. One single hive for the airlines would keep beehives interface clean and simple. 
Another upside would be that you'll only have to trigger one action to fetch data from the reqistered airlines and you'll also only have to evaluate one particular event. Then we could sort and evaluate most data inside of beehive and implement additional values into the events (f.e. `min_costs` or ordering flights per schedule) - for users comfort :)

muesli:
Agreed, a meta flight-search API might be more appealing.

Issue #89, opened 2017-02-15
Weather Bee


seblw:
A bee that provides weather information for given location etc.
Might use e.g. https://github.com/briandowns/openweathermap

PenguWin:
I'm on it :)

penguwin:
Merged in master

PR #173, opened 2017-10-08
Openweathermapbee

penguwin:
Now we're able to interact with some parts of the [openweathermaps](https://www.openweathermap.org) API
in beehive. This will enable us to fetch current weather information data.

Also see: #89


bees/openweathermapbee/event.go
```diff
@@ -0,0 +1,213 @@
...
+				Value: mod.current.Wind.Deg,
+			},
+			{
+				Name:  "clouds_all",
+				Type:  "int",
+				Value: mod.current.Clouds.All,
+			},
+			{
+				Name:  "rain",
+				Type:  "map[string]float64",
```
muesli: This probably isn't a supported type, yet. I'll have to investigate.

bees/openweathermapbee/openweathermapbeefactory.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+// Options returns the options available to configure this Bee.
+func (factory *OpenweathermapBeeFactory) Options() []bees.BeeOptionDescriptor {
+	opts := []bees.BeeOptionDescriptor{
+		{
+			Name:        "unit",
+			Description: "Your prefered unit",
+			Type:        "string",
+			Default:     "c", // celcius -> The right one
```
muesli: I laughed!

bees/openweathermapbee/event.go
penguwin: Well `rain`  and `snow` will return an emtpy map - the types are not supported yet, as supposed. Chains otherwise will continue working even if they access the empty map fields (tested with an ircbee).
We can think about keeping them inside for now and close the issue through supporting the type. If desired I can also comment-out those parts quickly ;)

PR #179, opened 2017-12-19
Add CANONICAL_URL env parameter for an container

leopku:
In a docker container, canonical url of beehive was not configurable and always act as http://localhost:8181. This pr accepting canonical url from env and passing it to beehive when beehive starting.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -17,5 +17,4 @@ WORKDIR /go/src/github.com/muesli/beehive
 VOLUME /conf
 
 # This form of ENTRYPOINT allows the beehive process to catch signals from the `docker stop` command
-ENTRYPOINT ["beehive", "-config", "/conf/beehive.conf", "-bind", "0.0.0.0:8181", "-canonicalurl", ${CANONICAL_URL}]
-
+ENTRYPOINT beehive -config /conf/beehive.conf -bind 0.0.0.0:8181 -canonicalurl ${CANONICAL_URL}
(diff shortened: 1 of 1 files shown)


muesli:
Nice, thank you for the contribution!

docker/Dockerfile
```diff
@@ -17,5 +17,4 @@ WORKDIR /go/src/github.com/muesli/beehive
 VOLUME /conf
 
 # This form of ENTRYPOINT allows the beehive process to catch signals from the `docker stop` command
-ENTRYPOINT ["beehive", "-config", "/conf/beehive.conf", "-bind", "0.0.0.0:8181", "-canonicalurl", ${CANONICAL_URL}]
-
+ENTRYPOINT beehive -config /conf/beehive.conf -bind 0.0.0.0:8181 -canonicalurl ${CANONICAL_URL}
```
muesli: I'm worried that changing this form invalidates what the comment above states. Any insight on this?

WangQiru: `docker stop` will always shutdown the container, changing `ENTRYPOINT` will not affect the container to catch stop signal.

leopku: LGTM.
But `ENTRYPOINT ["beehive", "-config", "/conf/beehive.conf", "-bind", "0.0.0.0:8181", "-canonicalurl", ${CANONICAL_URL}]` works well for me though it did not been tested in many cases.

Issue #180, opened 2018-01-31
dockerized Beehive doesn't write into beehive.conf


HenryQW:
Creating bees in Beehive container doesn't write the conf into beehive.conf, therefore `/v1/bees` will always return an empty json of `{"bees":[]}`

This causes redirection to a blank page with console error when user has:

1. Clicked "Create Bee" on any creation page
2. Clicked any Bee on `chains/new/pickbee` page

<img width="476" alt="screen shot 2018-01-31 at 20 49 52" src="https://user-images.githubusercontent.com/5896343/35646805-4f701f9a-06c8-11e8-9deb-12fdeaaecee3.png">

sgama:
I believe this is still an issue.

I just pulled the latest image, tried to add both a Github Bee and a Twitter Bee and my page
```
/bees/all
```
Shows empty.

PR #181, opened 2018-01-31
Alpine based Beehive docker image

WangQiru:
Squashed docker image size from 802MB to 523MB.


WangQiru:
Note: this does not fix #180.

muesli:
Thanks for your contribution! I'm currently looking into #180, should hopefully have a fix soon.

PR #182, opened 2018-02-03
Remove duplicate code in Dockerfile

WangQiru:


docker/Dockerfile-alpine
```diff
@@ -7,7 +7,6 @@ ENV URL=http://localhost:8181
 
 # Install dependencies and beehive
 RUN apk add --no-cache git build-base --no-cache && \
-    go get github.com/muesli/beehive && \
```
WangQiru: Duplicated.

muesli:
Should have caught this myself. Thanks for the update!

PR #197, opened 2018-09-30
update Gopkg.lock (go-mastodon & linkheader were missing)

Horrendus:
Updated Gopkg.lock


muesli:
Thanks for the contribution, but as discussed a while ago we finally made the switch to Go modules. No more Gopkg.lock, hence.

PR #199, opened 2018-10-10
Mixcloud bee

Horrendus:
MixcloudBee that can follow a Mixcloud user/feed and creates an event when a new CloudCast was uploaded

As it is one of my first serious go & beehive contributions I leave it atm as WIP for review. Please also review the new dependency https://github.com/Horrendus/go-mixcloud


muesli:
Nice work! The first thing that pops out when looking at go-mixcloud, is that you moved the actual code in a sub-dir (mixcloud) and separated the tests into their own package.

This can be done (especially if you want to write your tests like an external package user), but the common approach is to move them along side the actual code (so you would have `somefile.go` and `somefile_test.go`. That way godoc can also extract examples from your docs.

Either way, I'd recommend moving the actual code directly into the go-mixcloud directory. Each Go package is only composed of one directory, so a subdirectory `tests` wouldn't mess things up.

Horrendus:
I think I applied all your review comments on go-mixcloud, the mixcloud bee is not yet updated (need to read about go modules :) ).

For further review of the dependency I opened this issue here: https://github.com/Horrendus/go-mixcloud/issues/1

Further review of the mixcloud bee should of course still be done here once I updated it ;)

muesli:
The good news is: Go modules doesn't change much for you right now.

Make sure to use Go 1.11+, and either set GO111MODULE=on, or compile beehive outside your $GOPATH. This will activate Go modules support. Now, if you import go-mixcloud as a dependency in your bee, Go will automatically detect that and update the go.mod file for you. Make sure to include the changes to that file in your pull request.

"The go command resolves imports by using the specific dependency module versions listed in go.mod. When it encounters an import of a package not provided by any module in go.mod, the go command automatically looks up the module containing that package and adds it to go.mod, using the latest version"

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,124 @@
...
+						Value: cloudcast.URL,
+					},
+					{
+						Name:  "slug",
+						Type:  "string",
+						Value: cloudcast.Slug,
+					},
+				},
+			}
+			mod.Logln("Event new_cloudcast", newCloudcastEvent)
```
muesli: I'd refrain from printing this out manually here, as every occurring event will eventually be printed out to stdout regardless.

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,124 @@
...
+			mod.eventChan <- newCloudcastEvent
+		}
+		time.Sleep(25 * time.Second)
```
muesli: Any good reason to do this every 25 seconds? :smile:

muesli:
Looking good!

bees/mixcloudbee/mixcloudbee.go
Horrendus: Yeah to continuosly get new CloudCasts that might have been added later, but I guess that should/could be done by a CronBee in Front of it :)

Then I guess it would just need to persist the lastUpdate and have an option "sinceLastUpdateOnly" in case the user wants to ignore the lastUpdate value.

Horrendus:
The Bee no longer sleeps, it now has an action "pollFeed" that can be triggered by e.g. a CronBee. Not fully sure if I correctly did the options part of the action (Action shouldnt have any option).

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,133 @@
...
+
+	client     *mixcloud.Client
+	lastUpdate time.Time
+
+	eventChan chan bees.Event
+}
+
+// Poll a Mixcloud Cloudcasts feed
+func (mod *MixcloudBee) pollFeed(feed string) {
+	mod.Logln("Parsing feed")
```
CalmBit: A _bit_ nitpicky, but this logging statement could probably be better served with a mod.LogDebugf or similar, as to not clutter the main output. @muesli can confirm whether or not I'm right in making this assessment 😄

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,133 @@
...
+	var allCloudcastsData []mixcloud.CloudcastData
+	var opt mixcloud.ListOptions
+	opt.Since = mod.lastUpdate
+	opt.Until = time.Now()
+	mod.lastUpdate = opt.Until
+	mod.Logln("Since", opt.Since, "Until", opt.Until)
```
CalmBit: Ditto here 😄

Horrendus:
@CalmBit: Thanks for the suggestions!

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	return outs
+}
+
+// Run executes the Bee's event loop.
+func (mod *MixcloudBee) Run(cin chan bees.Event) {
+	mod.eventChan = cin
+
+	time.Sleep(10 * time.Second)
```
muesli: We can remove this line now :smile:

bees/mixcloudbee/mixcloudbee.go
```diff
@@ -0,0 +1,133 @@
...
+		mod.eventChan <- newCloudcastEvent
+	}
+}
+
+// Action triggers the action passed to it.
+func (mod *MixcloudBee) Action(action bees.Action) []bees.Placeholder {
+	outs := []bees.Placeholder{}
+
+	switch action.Name {
+	case "pollFeed":
```
muesli: Let's stick with the scheme here and call it `poll_feed`.

bees/mixcloudbee/mixcloudbeefactory.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	return events
+}
+
+// Actions describes the available actions provided by this Bee.
+func (factory *MixcloudBeeFactory) Actions() []bees.ActionDescriptor {
+	actions := []bees.ActionDescriptor{
+		{
+			Namespace:   factory.Name(),
+			Name:        "pollFeed",
```
muesli: Let's stick with the scheme here and call it `poll_feed`.

muesli:
LGTM! Just waiting for TravisCI to finish.

PR #200, opened 2018-10-17
Fixed bad parse query args

dgrr:


muesli:
Could you elaborate what exactly was broken before?

dgrr:
The error occurrs when you use query_params variable in a template (in this case I use `{{index .query_params "id"}}`). Request URI query is wrong parsed.
Basing on [rfc3986](https://tools.ietf.org/html/rfc3986#section-3) the Request URI is something like 
```[scheme:][//[userinfo@]host][/]path[?query][#fragment]```

If you make a request to `http://host:port/?id=90` the query parsed to [url.ParseQuery](https://godoc.org/net/url#ParseQuery) is `/?id=90` which in Go map will be `map["/?id"] = 90`.
So the value of `map["id"]` is missing because of the bad value parsing.

This commit tries to fix URI query parsing.

@muesli

muesli:
I see, that makes sense. Thanks for your contribution, much appreciated!

Issue #205, opened 2018-10-21
Improve Dockerfile


Coksnuss:
I found this repository today and wanted to try beehive. Unfortunately the Dockerfile seems broken as the `go get` command fails. I then tried the image that is hosted on Docker Hub and found that it is over two years old and contains an old binary of beehive.

Because of that I created my own Dockerfile for beehive. It uses the Docker multi-stage build feature and results in a MUCH smaller image (24.2 MB compared to 800 MB). Beehive is build with the `embed` option which makes it possible to use a single binary within an alpine container.

Feel free to integrate this file to the repository or provide it as an additional alternative to the existing ones.

```
FROM golang:1.11-alpine

RUN apk add -U --no-cache curl git make g++ && \
    curl "https://raw.githubusercontent.com/golang/dep/master/install.sh" | sh && \
    go get -d -v github.com/muesli/beehive && \
    cd ${GOPATH}/src/github.com/muesli/beehive && \
    make get-deps && \
    make


FROM alpine:3.8

RUN apk add -U --no-cache ca-certificates

WORKDIR /app

COPY --from=0 /go/src/github.com/muesli/beehive/beehive .

ENTRYPOINT ["./beehive"]
```

HenryQW:
LGTM, the Dockerfile is indeed out-dated and I haven't got time to update it, please PR this!

Coksnuss:
> You can use `scratch` instead of `alpine:3.8` to make it even smaller.

Good point but it refuses to run if i use `scratch`:

> standard_init_linux.go:190: exec user process caused "no such file or directory"

Furthermore it wouldn't contain the `ca-certificates` package which is important for crawling RSS feeds over HTTPS. Otherwise fetching would fail with

> ERRO[0085] [Fefe Blog]: https://blog.fefe.de/rss.xml: Get https://blog.fefe.de/rss.xml: x509: certificate signed by unknown authority

muesli:
I've pushed some updates to the `Dockerfile` in git master, and new docker images will from now on be automatically built on the docker hub.

PR #206, opened 2018-10-27
Add Dockerfile that creates a low-footprint image

schnz:
I moved the "old" Dockerfiles to a `legacy` directory and replaced them with the new Dockerfile (See #205 for more information).

Also rewrote the README.md


docker/Dockerfile
```diff
@@ -1,20 +1,20 @@
...
+WORKDIR /app
 
-# create a volume for the configuration persistence
-VOLUME /conf
+COPY --from=0 /go/src/github.com/muesli/beehive/beehive .
 
-# This form of ENTRYPOINT allows the beehive process to catch signals from the `docker stop` command
-ENTRYPOINT beehive -config /conf/beehive.conf -bind 0.0.0.0:8181 -canonicalurl ${CANONICAL_URL}
+ENTRYPOINT ["./beehive"]
+CMD ["-bind", "0.0.0.0:8181", "-canonicalurl", "http://localhost:8181"]
```
muesli: I think it would be nice to keep the canonicalurl an ENV var. This is essentially a user setting.

Coksnuss: You are right. Thats why these parameters are specified as `CMD` instead of `ENTRYPOINT` so that the user can override them when initially creating the container (also see README.md).

muesli: Interesting, I didn't know you could do that. I still wonder if that's common practice though? I've so far only seen projects using the ENV approach.

muesli:
I've now updated the Dockerfile with a bunch of your suggestions. Sorry it took that long, but I really wanted to make the switch to Go modules first.

I'd love hear your feedback on the current situation, though!

Coksnuss:
I would still use a combination of ENTRYPOINT and CMD in order to let the user customize the parameters that are being passed to beehive or at least favor CMD over ENTRYPOINT.

> I still wonder if that's common practice though?

I don't know if there is a 'common practice' with that. I think it highly depends on the use case.
Some images give the user total freedom by not specifying an entrypoint at all.
```
CMD ["nginx", "-g", "daemon off;"]
```

Other images allow the user to pass arbitrary parameters without supplying a default parameter
```
ENTRYPOINT [ "certbot" ]
```

I think for beehive it makes sense to have a default command that can be overridden by the user if he likes to change the listening interface (e.g. when using the `--network host` switch)

Issue #204, opened 2018-10-21
make get-deps: No versions of github.com/google/go-github met constraints


LorenzoAncora:
Installation of the latest version fails with this error:

$ make get-deps
> dep ensure
Solving failure: No versions of github.com/google/go-github met constraints:
        v15.0.0: Could not introduce github.com/google/go-github@v15.0.0, as its subpackage github.com/google/go-github/v18/github is missing. (Package is required by (root).)
        v18.2.0: Could not introduce github.com/google/go-github@v18.2.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v18.1.0: Could not introduce github.com/google/go-github@v18.1.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v18.0.0: Could not introduce github.com/google/go-github@v18.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v17.0.0: Could not introduce github.com/google/go-github@v17.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v16.0.0: Could not introduce github.com/google/go-github@v16.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v14.0.0: Could not introduce github.com/google/go-github@v14.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v13.0.0: Could not introduce github.com/google/go-github@v13.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v12.0.0: Could not introduce github.com/google/go-github@v12.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v11.0.0: Could not introduce github.com/google/go-github@v11.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v10.0.0: Could not introduce github.com/google/go-github@v10.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v9.0.0: Could not introduce github.com/google/go-github@v9.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v8.0.0: Could not introduce github.com/google/go-github@v8.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v7.0.0: Could not introduce github.com/google/go-github@v7.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
        v6.0.0: Could not introduce github.com/google/go-github@v6.0.0, as it is not allowed by constraint ^15.0.0 from project beehive.
[...]

PR #207, opened 2018-11-04
Correct installation instructions

LorenzoAncora:
Complete installation instructions (with comments for inexperienced users), enhanced semantic formatting, important installation notes, security notes and interactive anchors for the "Troubleshooting & Notes" section.

This should facilitate the installation procedure and avoid forgetfulness by the end user.

Mitigates #204

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,8 +6,8 @@ agents that perform automated tasks triggered by events and filters. It is
 modular, flexible and really easy to extend for anyone. It has modules
 (we call them *Hives*), so it can interface with, talk to, or retrieve
 information from Twitter, Tumblr, Email, IRC, Jabber, RSS, Jenkins, Hue - to
-name just a few. Check out the full list of [available Hives](https://github.com/muesli/beehive/wiki/Available-Hives)
-in our Wiki.
+name just a few. Check out the full list of
+[available Hives](https://github.com/muesli/beehive/wiki/Available-Hives) in our Wiki.
 
 Connecting those modules with each other lets you create immensly useful agents.
 
@@ -23,13 +23,23 @@ Connecting those modules with each other lets you create immensly useful agents.
 
 ## Installation
 
-Beehive requires Go 1.8 or higher. Make sure you have a working Go environment. See the [install instructions](http://golang.org/doc/install.html).
+Beehive requires Go 1.8 or higher. Make sure you have a working Go environment.
+See the [installation instructions of Go](http://golang.org/doc/install.html).
 
 ### From source
 
-The recommended way is to fetch the sources and run make. This requires [dep](https://github.com/golang/dep)
-to be installed on your system.
+The recommended way is to fetch the sources and run make.
+This requires [dep](https://github.com/golang/dep) to be installed on your system.
 
+If the `$GOPATH` environment variable is empty (`echo $GOPATH`), set it with:
+    
+    # Set the $GOPATH environment variable for the active shell
+    export GOPATH="/your/go/path"
+
+Create the directory tree `$GOPATH/src/github.com/muesli/` and run this script:
+
+    cd $GOPATH/src/github.com/muesli/
+    # GNU Git should download the sources in the beehive/ directory
     git clone https://github.com/muesli/beehive.git
     cd beehive
     make get-deps
@@ -37,9 +47,10 @@ to be installed on your system.
 
 You can build and install the `beehive` binary like other Go binaries out there (`go get`)
 but you'll need to make sure Beehive can find the assets (images, javascript, css, etc).
-See the Troubleshooting/Notes section for additional details.
+See the [Troubleshooting & Notes](#TAndN) section for additional details.
 
-Run `beehive --help` to see a full list of options.
+Run `beehive --help` to see a full list of options (in particular to set the
+access port of the administrative interface and the URL from which you can visit Beehive).
 
 ### Deployment Tools
  - [Dockerfile](docker)
@@ -73,9 +84,9 @@ in your browser. Note that Beehive will create a config file `beehive.conf`
 in its current working directory, unless you specify a different file with the
 `-config` option.
 
-Note: You currently have to start `beehive` from within $GOPATH/src/github.com/muesli/beehive
-in order for it to find all the resources for the admin interface. Also see the
-Troubleshooting & Notes section of this README.
+**Note**: you currently have to install Beehive in `$GOPATH/src/github.com/muesli/beehive`
+and start `beehive` from there in order for it to find all the resources for the admin interface.
+Also see the [Troubleshooting & Notes](#TAndN) section of this README.
 
 The admin interface will present you with a list of available Hives. We will
 need to create two Bees here, one for the RSS feed and one for your email
@@ -108,7 +119,10 @@ That's it. Whenever the RSS-feed gets updated, Beehive will now send you an
 email! It's really easy to make various Bees work together seamlessly and do
 clever things for you. Try it yourself!
 
-You can find more information on how to configure Beehive and examples [in our Wiki](https://github.com/muesli/beehive/wiki/Configuration).
+You can find more information on how to configure Beehive and examples
+[in our Wiki](https://github.com/muesli/beehive/wiki/Configuration).
+
+<a name="TAndN"></a>
 
 ## Troubleshooting & Notes
 
@@ -121,12 +135,15 @@ Should you still not be able to reach the web interface, check if the `config`
 directory in the git repository is empty. If that's the case, make sure the
 git submodules get initialized by running `git submodule update --init`.
 
-The web interface does *not* require authentication yet. Beehive currently
-accepts all connections from the loopback device *only*.
+The web interface does *not* require authentication yet. Beehive by default
+accepts all connections from the loopback device *only* (if you allow remote
+access, **do not forget to configure the firewall** to allow access only from trusted hosts).
 
 ## Development
 
-Need help? Want to hack on your own Hives? Join us on IRC (irc://freenode.net/#beehive) or [Gitter](https://gitter.im/the_beehive/Lobby). Follow the bees on [Twitter](https://twitter.com/beehive_app)!
+Need help? Want to hack on your own Hives? Join us on IRC (irc://freenode.net/#beehive)
+or [Gitter](https://gitter.im/the_beehive/Lobby).
+Follow the bees on [Twitter](https://twitter.com/beehive_app)!
 
 [![GoDoc](https://godoc.org/github.com/golang/gddo?status.svg)](https://godoc.org/github.com/muesli/beehive)
 [![Build Status](https://travis-ci.org/muesli/beehive.svg?branch=master)](https://travis-ci.org/muesli/beehive)


README.md
```diff
@@ -23,23 +23,34 @@ Connecting those modules with each other lets you create immensly useful agents.
...
 
+If the `$GOPATH` environment variable is empty (`echo $GOPATH`), set it with:
+    
+    # Set the $GOPATH environment variable for the active shell
+    export GOPATH="/your/go/path"
+
+Create the directory tree `$GOPATH/src/github.com/muesli/` and run this script:
+
+    cd $GOPATH/src/github.com/muesli/
+    # GNU Git should download the sources in the beehive/ directory
```
muesli: I find this comment to be a bit superfluous.

README.md
```diff
@@ -23,23 +23,34 @@ Connecting those modules with each other lets you create immensly useful agents.
...
 
-The recommended way is to fetch the sources and run make. This requires [dep](https://github.com/golang/dep)
-to be installed on your system.
+The recommended way is to fetch the sources and run make.
+This requires [dep](https://github.com/golang/dep) to be installed on your system.
 
+If the `$GOPATH` environment variable is empty (`echo $GOPATH`), set it with:
+    
+    # Set the $GOPATH environment variable for the active shell
+    export GOPATH="/your/go/path"
```
muesli: The most recent Go versions don't actually require you to set this env variable: if it's not set, it defaults to `$HOME/go`. Maybe we should consider that the default in the install instructions, too.

README.md
```diff
@@ -23,23 +23,34 @@ Connecting those modules with each other lets you create immensly useful agents.
...
+
+Create the directory tree `$GOPATH/src/github.com/muesli/` and run this script:
```
muesli: I'd just say "...and run:" - it's not really a script after all.

README.md
```diff
@@ -23,23 +23,34 @@ Connecting those modules with each other lets you create immensly useful agents.
...
+
+    cd $GOPATH/src/github.com/muesli/
+    # GNU Git should download the sources in the beehive/ directory
     git clone https://github.com/muesli/beehive.git
     cd beehive
     make get-deps
```
muesli: Before `make get-deps` we should probably add a `git submodule update --init` just to be safe.

README.md
```diff
@@ -73,9 +84,9 @@ in your browser. Note that Beehive will create a config file `beehive.conf`
 in its current working directory, unless you specify a different file with the
 `-config` option.
 
-Note: You currently have to start `beehive` from within $GOPATH/src/github.com/muesli/beehive
-in order for it to find all the resources for the admin interface. Also see the
-Troubleshooting & Notes section of this README.
+**Note**: you currently have to install Beehive in `$GOPATH/src/github.com/muesli/beehive`
```
muesli: That's not actually true: you can install Beehive from any directory, if you know what you're doing. The important part is running it from the folder with the checkout git submodules, so it can find the HTML/JS sources.

README.md
LorenzoAncora: @muesli the Beehive license is AGPL, so the user can get the source from anyone without the need to use Git and potentially without having seen this repository and its Issues section.
We must consider the worst case: the end user does not know how Git works, so he must be informed that he is going to run GNU Git (in this way he can install it if he is missing) and that his execution _must_ produce a directory with the sources.

README.md
LorenzoAncora: @muesli this is false on my version of Debian ARM64 (the most recent). The user must explicitly set GOPATH. This also happens with Python and the other interpreters. This is good: for security reasons the *PATH variables should be explicitly set by the end user, possibly inheriting the defaults set explicitly by the administrator in compliance with SELinux configurations and other security systems.

README.md
LorenzoAncora: @muesli a script is by definition a sequence of two or more instructions, regardless of whether they are in a file or not.

README.md
LorenzoAncora: @muesli the Makefile already contains:
```
submodule:
     git submodule update --init
```
It is already automatically executed with the simple "make" command, in order before the other commands. If you think it's a necessary command then add it, but more commands are always equivalent to less users willing to run them.

README.md
LorenzoAncora: @muesli the user can also rewrite Beehive from scratch, if he knows what he is doing. :-) If I had difficulty and I have been using GNU Git for years, I think it is appropriate to give simple instructions: if the end user "knows what he does", then he will guess that he can make changes.
As always in programming the worst case prevails: the user does not know how git works or how Beehive's dependencies work.

README.md
muesli: Which version of Go does it come with? Beehive depends on 1.8, the very version that made GOPATH optional and not required.

README.md
muesli: Sorry, I really don't want extra comments in the commands section here. If you really consider it that important, it should be mentioned above in the "regular" README text.

Personally I still think it's a bit silly: if you're trying to compile a piece of software yourself and haven't stumbled upon git yet, you're in for a ride. We don't explain the expected behavior of "cd" and "export" here either.

README.md
muesli: You're right!

README.md
LorenzoAncora: > Which version of Go does it come with? Beehive depends on 1.8, the very version that made GOPATH optional and not required.

The version of the Debian package is: `2:1.10~5`.
The version of the Go interpreter is: `go version go1.10.4 linux/arm64`.
@muesli I am afraid that regardless of the version you will have to adapt to the concept that the environment variables are subject to security measures and must therefore be set explicitly. For example, in my system GOPATH must be set first in the shell during installation and then in the systemd service, because systemd isolates the program from the environment (always a good thing) and therefore you must always explicitly pass the environment variables. :-)
In this regard after you have accepted this push I will correct the systemd service file so that it can work without any further changes on all Debian systems. ;-)

README.md
LorenzoAncora: > Sorry, I really don't want extra comments in the commands section here

@muesli a code without comments is always a badly written code.

> We don't explain the expected behavior of "cd" and "export" here either.

Of course, but you are confusing external commands with built-in shell commands and system tools.
In all systems the built-ins are documented by the shell itself and therefore the user should know them; the system tools on GNU/Linux are the GNU utilities, which you can take for granted the user knows how to use (they are present on all systems and well documented, so in one way or another the user manages to use them).
Git, however, is an external command (it is not a GNU system tool, but an optional GNU software) and therefore you can not (_and must not_) make assumptions about the fact that the user has the documentation and should know how it works (so to speak, the "man" is totally optional and only a small slice of users know how it works git in detail).
As a developer it's your job to consider the worst case: the user wants to use the software and the compilation is just a hassle to accomplish this end. If put in front of a fork in the bad case the user always chooses the solution ready to use.

**If you can not provide a ready-to-use system package (means "apt install beehive" and then it works), then you must make the installation intuitive for all users, regardless of their degree of experience.**

README.md
LorenzoAncora: @muesli the push request #209 completes the facilitation work by providing a clearer and more friendly installer. In this way even if they (the end users) have read the documentation in a superficial way they can understand all the changes made to the system.
I used simple syntaxes to allow anyone to change the installation process over time. In addition I have also added the classic "contributors" file to attract other developers. Obviously you will have to manually give approval for this commit too.

README.md
LorenzoAncora: @muesli then? It's been 2 days, do you want to merge or not?

LorenzoAncora:
@muesli then? It's been 5 days, do you want to merge or not? :-)
Remember that you are free to add as many commits as you want to the push request.

LorenzoAncora:
> Sorry, I've been a little under the weather this week. I'll make a few changes and merge this tonight. 

I suggest you _add_ comments. ;-)

> Thanks for your contributions!

@muesli you're welcome.

LorenzoAncora:
> [3 days ago] I'll make a few changes and **merge this tonight**.

@muesli forgot anything? :-)

LorenzoAncora:
@muesli ...are you alive? :'-)

README.md
rubiojr: @LorenzoAncora given that `GOPATH` is now optional in go 1.8+ and set to `~/go` by default, I'm wondering if we should just tell people to use `~/go` to simplify the instructions and avoid the export.

prologic: We should instead just upgrade the deps and docs  to be compatible with Go11Modules.  It's far easier and  just works. One does not need  to care about  any `$GOPATH`.

muesli:
Cherry-picked a bunch of your suggestions into f5e968d0407c6903f1481879ec329a728528ce8c.

Thanks for your contribution - sorry it took so incredibly long. Now that we've finally switched to Go modules we can start cleaning up a bit.

LorenzoAncora:
> Cherry-picked a bunch of your suggestions into f5e968d.

@muesli great.
Please, could you add the _Merged_ label to this Pull request? :-)

LorenzoAncora:
@muesli great.
Please, could you add the _Merged_ label to this PR? :-)

muesli:
@LorenzoAncora I don't think you can change that manually - or at least I wouldn't know how to.

LorenzoAncora:
>  Please, could you add the _Merged_ **label** to this PR? :-)
>> @LorenzoAncora I don't think you can change that manually - or at least I wouldn't know how to.

@muesli, the GitHub guide explains how to assign [create](https://help.github.com/en/articles/creating-a-label) and [assign](https://help.github.com/en/articles/applying-labels-to-issues-and-pull-requests?query=create%20label) a label.

muesli:
@LorenzoAncora You're talking about the read "Closed" and purple "Merged" though, no? Those aren't labels.

LorenzoAncora:
> Please, could you add the Merged **label** to this PR? :-)
> @muesli, the GitHub guide explains how to **[create](https://help.github.com/en/articles/creating-a-label) and [assign](https://help.github.com/en/articles/applying-labels-to-issues-and-pull-requests?query=create%20label) a label**.
>> @LorenzoAncora I don't think you can change that manually - or at least I wouldn't know how to.
>> You're talking about the read "Closed" and purple "Merged" though, no? Those aren't labels.

@muesli no, I am not talking about the status, I am talking only about labels.
I suggested you add the _homonymous_ **label** "Merged", to signify that the status Closed is not true.

[](https://help.github.com/en/articles/creating-a-label)

LorenzoAncora:
@muesli ...**are you really experiencing serious difficulties in completing such a simple operation**?
I see you as very active, a couple more clicks shouldn't break you too much.

muesli:
@LorenzoAncora Sorry, but that's not the way we use labels and there's simply no other way to mark a pull request as merged manually. Is this really such a big issue?

LorenzoAncora:
> Sorry, but that's not the way we use labels

@muesli why do you speak in plural if you are the only copyright holder and the repository administrator? If yours were an official team, there would be an `AUTHORS` or `CONTRIBUTORS` file with the list of members to be credited.
The decision on how to use the labels is uniquely yours. :-)

Labels were created specifically for these situations where Git does not suffice.

> Is this really such a big issue?

Imagine if someone behaved the same way with all your commits and they all looked rejected while they were actually integrated. It certainly isn't _nice_ to see.
It's a matter of courtesy, it should be _spontaneous_.

muesli:
> @muesli why do you speak in plural if you are the only copyright holder and the repository administrator? If yours were an official team, there would be an `AUTHORS` or `CONTRIBUTORS` file with the list of members to be credited.

There's been a total of 34 contributors to Beehive. If you look at the sources you'll notice a bunch more copyright messages. There's no such thing as an "official team" when it comes to open source development. There may be core developers who commit on a more regular basis, but even then `AUTHORS` and `CONTRIBUTORS` files became obsolete ever since we switched to using version control systems. You can simply look that info up in the repository or on GitHub.

> The decision on how to use the labels is uniquely yours. :-)

Yes, and as such I would appreciate if you respect the way we/I handle labels.

> Imagine if someone behaved the same way with all your commits and they all looked rejected while they were actually integrated. It certainly isn't _nice_ to see.

It's not that hard to imagine, as it's exactly what regularly happens when you contribute to various projects, with a slightly different development routine. Go (the language & tool-set) itself among them. One of my commits as an example: https://github.com/golang/tools/pull/142

I can understand open requests lingering around being a bit bothersome, but that's no the case anymore.

LorenzoAncora:
> [...] **AUTHORS and CONTRIBUTORS files became obsolete** ever since we switched to using version control systems. You can simply look that info up in the repository or on GitHub. [...]

Since you don't know where your repository will end up over the years and you don't want to be a victim of vendor lock-in, you _should_ always maintain a file giving credit to contributors in chronological order. It is simply good practice, not an imposition and it is also _not obsolete_.

> **There's been a total of 34 contributors** to Beehive. [...]

There is no shame in developing software only with contributions from passing users, even if it is such a small number. Every contribute is precious. :-)
You are the only administrator and the only copyright holder of the software, because at the moment it is not stated otherwise in the `README` or in the `COPYING`\\`LICENSE` files: unless you always refer to yourself as "we", when not having an official team or a dedicated company you shouldn't say "we" when it comes to administration.

> The decision on how to use the labels is uniquely yours. :-)
>> [...] Yes, and as such I would appreciate if you **respect the way we/I handle labels**. [...]

Whether I respect it or not does not make any difference: you are the only one who imposes labels, other users can only adapt. I asked you to put a label because I think it's the right thing to do.

> Imagine if someone behaved the same way with all your commits and they all looked rejected while they were actually integrated. It certainly isn't nice to see.  It's a matter of courtesy, it should be spontaneous.
>> [...] It's **not that hard to imagine, as it's exactly what regularly happens** when you contribute to various projects, with a slightly different development routine. [...]

Actually in many years you are the only one with whom I had this problem.
On the other hand, GitHub has a labeling system to overcome these mistakes without altering the Git metadata: a couple of clicks would have been enough to solve the problem.

[muesli -> unsubscribe]()

Now I have work to do.

PR #209, opened 2018-11-05
 Enhanced Makefile with realtime comments and contributors file

LorenzoAncora:
This new Makefile explains the actions performed (and their outcome) in real time, as a business grade installer would do.
It is designed in such a way as to give constant feedback to the user through standard GNU Make syntaxes and provide suggestions where appropriate without being overly verbose. [...]

For more informations please read the commits.

The purpose of this push request is to increase the number of users and attract new developers by enhancing the UX.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -1,0 +1,15 @@
+This is the list of all the people that make this great software.
+Add yourself at the bottom of the list if you do contribute and wish
+to appear as a contributor. At the very minimum, the GitHub username
+is required if you do wish to appear. Other fields are optional.
+
+Format: Firstname Lastname @github-handle <email>
+
+CONTRIBUTORS
+------------
+
+Christian Muehlhaeuser @muesly <muesli@gmail.com>
+Michael Wendland @michiwend
+Gabriel Alacchi @GabrielAlacchi
+Johannes Fürmann @waaaaargh <johannes@fuermann.cc>
+Lorenzo L. Ancora @LorenzoAncora <admin@lorenzoancora.info>
--- a/Makefile
+++ b/Makefile
@@ -4,30 +4,62 @@ COMMIT_SHA=$(shell git rev-parse --short HEAD)
 all: embed
 
 submodule:
+	@echo "Downloading and updating Beehive's submodules..."
+	@echo -n "...executing: "
 	git submodule update --init
+	@echo "Done."
 
 noembed: submodule build
+	@echo "Resource embedding: disabled.\n\
+	You can use 'make embed' to enable it."
 
 generate:
+	@echo "Embedding binary data in go files..."
+	@echo -n "...executing: "
 	$(shell go env GOPATH)/bin/go-bindata --tags embed --pkg api -o api/bindata.go --ignore \\.git assets/... config/...
+	@echo "Data embedded."
 
 go-bindata:
-	[ -f $(shell go env GOPATH)/bin/go-bindata ] || go get -u github.com/jteeuwen/go-bindata/go-bindata
+	@echo -n "Management of go-bindata..."
+	@if [ ! -f $(shell go env GOPATH)/bin/go-bindata ]; then\
+		GOBINDATA_GET="go get -u github.com/jteeuwen/go-bindata/go-bindata";\
+		echo "\nDownloading go-bindata for resource embedding...";\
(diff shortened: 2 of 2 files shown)


LorenzoAncora:
@muesli are you alive? :-)

muesli:
We do so little in the Makefile now I think we should just stick to its default behavior. It already prints out what it's going to call/do.

LorenzoAncora:
> We do so little in the Makefile now I think we should just stick to its default behavior. It already prints out what it's going to call/do.

@muesli I know, but that's the point: the problem is that the verbose output makes it extremely difficult to understand what is happening. The make system was designed specifically to allow this type of customization and, by default, produces information suitable for parsing but not for reading. As the complexity of the makefile increases (and inevitably will happen) the problem gets worse.
An end-user friendly installer will make the application more popular.
Are you still convinced that rejecting this commit is beneficial?

CalmBit:
By and large, if we're looking for a user-friendly installer, we will more than likely package the system and the binaries in a format that doesn't require building everything. That's the end goal, as far as I understand it - binary distribution through better channels than a makefile. Developers, on the other hand, don't necessarily need all the bells and whistles here.

LorenzoAncora:
> [...] end goal [...] **binary distribution** through better channels than a makefile [...] Developers, on the other hand, don't necessarily need all the bells and whistles here.

The distribution of only executable components is typical of Microsoft Windows systems, while most GNU / Linux users install open source software directly from tarballs. I anticipate that if you are alone in managing the project and you do not have paid employees, treating and distributing binary versions with installers for each release is practically impossible, especially if you want to distribute the software on multiple platforms. Judging by its structure (and therefore by the patterns used) Beehive is one of these software whose complexity grows rapidly over time and requires frequent updates, so you will inevitably end up distributing tarballs.
Then, users who want the latest version with the latest changes will still clone the repository (which is equivalent to getting a tarball) and will use the makefile.
I've proposed the change because, according to my experiences, the free softwares compatible with GNU / Linux always end up depending on makefiles, which is a good thing, because it allows everyone to analyze and customize the installation system.

CalmBit:
Not everyone uses tarballs, and there's more of a movement for general purpose computing to binary formats. Distributing binaries is _not_ difficult with Go - that's the fun part. Files like .deb packages and AUR PKGCONFIGs can be far more accessible and dependable than fickle makefiles made for developers who know what they're doing. If you're touching a makefile, it's likely you know what you're doing. If you don't know what you're doing, no amount of echo statements in the world will fix that.

LorenzoAncora:
> **fickle makefiles** [...] **If you don't know what you're doing, no amount of echo statements in the world will fix that.**

As administrator of the UNESCO GNU Free Software Directory, I personally followed the development of different applications and over time I got my very personal idea of how things end up. I am the administrator who proposed and approved the inclusion of Beehive in the directory, after an evaluation process on both the actual state of the software and its probable future. My interest is also to prevent obsolescence and facilitate the transition to GNU standards.
Makefiles are meant to be used by both end users and developers, and if they are designed correctly, they almost never create problems.
Having experience with GNU development environments, I can confidently say that customizing a makefile is always necessary to ensure a good UX. The syntax of GNU make is specifically designed to allow this kind of customization, because it should never be assumed that the end user is competent. GNU make is designed to guarantee backward compatibility over time and usability on different devices.
That being said, I strongly advise you not to seek independence from GNU make because it actually improves the quality of the software itself.

> **Not everyone uses tarballs**, and there's more of a movement for general purpose computing to binary formats. Distributing binaries is not difficult with Go - that's the fun part. [...]

From my point of view (which inevitably puts GNU/Linux first), all long-term users will eventually use the tarball, especially after the software enters the maintenance phase.
A functional and easy-to-use tarball is one of the best qualities that can be found in a free software package, obviously after proper licensing and documentation: users who want to understand how the software works tend to download the tarball of a specific version that they consider more stable (or in any case to clone the repository) and then proceed to make some changes and, optionally, to send the updated code as a patch or as a commit to the initial repository.
End users want a simple and clear indicator of what is happening and like easy-to-use installers. Simple modifications (yes, like echoes with some branch logic :-) ) are often the most effective and cause fewer problems over time.

You describe yourself as "programmer, musician, linguist, creator of things" and you seem really sure of what you do, so I am sure that other people will benefit more from my contributions and suggestions... bb :-)

Issue #212, opened 2018-12-25
can't build on head with go 1.11.4


ChangePlaces:
/root/go/src/github.com/muesli/beehive/api/resources/bees/bees_post.go:63:31: not enough arguments in call to smolder.ErrorResponseHandler
        have (*restful.Request, *restful.Response, *smolder.ErrorResponse)
        want (*restful.Request, *restful.Response, error, *smolder.ErrorResponse)
/root/go/src/github.com/muesli/beehive/api/resources/bees/bees_post.go:63:75: too many arguments in call to smolder.NewErrorResponse
        have (number, bool, error, string)
        want (int, interface {}, string)
# github.com/muesli/beehive/api/resources/chains
/root/go/src/github.com/muesli/beehive/api/resources/chains/chains_post.go:63:31: not enough arguments in call to smolder.ErrorResponseHandler
        have (*restful.Request, *restful.Response, *smolder.ErrorResponse)
        want (*restful.Request, *restful.Response, error, *smolder.ErrorResponse)
/root/go/src/github.com/muesli/beehive/api/resources/chains/chains_post.go:63:75: too many arguments in call to smolder.NewErrorResponse
        have (number, bool, string, string)
        want (int, interface {}, string)

rubiojr:
@ChangePlaces: https://github.com/muesli/beehive/pull/213 is trying to fix this.

ChangePlaces:
Thanks, would love to get this working on my pi!

muesli:
Should be fixed in master now. Note that we switched to Go modules.

PR #185, opened 2018-03-09
Use Gopkg.toml for dependencies

asymmetric:
See muesli/beehive-vendor#2 for a discussion.

PR #160, opened 2017-07-01
Tinderbee

penguwin:
Someone (@muesli) should really test this code

PR #213, opened 2019-04-03
Got build working again

cyian-1756:
I got the build working on Go 1.11.2 by removing 2 bees (tinderbee and slackbee) and making some minor changes so how error pages are called

The tinderbee was removed because the lib it used has been removed from github


rubiojr:
@cyian-1756 building fine here with:

```
[0][rubiojr@mantaray] go version
go version go1.12.1 darwin/amd64
```

What kind of error are you getting?

cyian-1756:
> What kind of error are you getting?

I was getting a 404 error on the tinderbee as the tinder lib it relied on has been pulled from github

The slackbee was throwing an error because the lib it uses has made breaking api changes (The exact error was about a wrong number of arguments being passed to postMessage)

The 2 other error messages I got were the same as in #212

rubiojr:
@cyian-1756 ah I see, it seems the way we vendor things changed in https://github.com/muesli/beehive/pull/185 where we stopped using muesli/beehive-vendor, so things will break if 3rd party libs disappear from the interwebs.

Perhaps @penguwin is interested in fixing the Tinder bee (per https://github.com/muesli/beehive/pull/160), instead of removing it, assuming that can be done. In any case, we should remove it from `Gopkg.toml` if we're 🔥 it.

You'll want to bump the slack dep in https://github.com/muesli/beehive/blob/master/Gopkg.toml#L138 to 0.5.0, otherwise the build will fail if we pull 0.2.

I'm not sure about the changes to `bees_post.go` and `chains_post.go`, but it seems there's a legit issue if you don't build against smolder `e6752d792454819bc63dec1a7a6f04bd7d539c7c` (current HEAD), so you'll want set that in `Gopkg.toml`, having master there is not enough:

```
api/resources/bees/bees_post.go:63:31: too many arguments in call to smolder.ErrorResponseHandler
	have (*restful.Request, *restful.Response, error, *smolder.ErrorResponse)
	want (*restful.Request, *restful.Response, *smolder.ErrorResponse)
api/resources/bees/bees_post.go:63:80: not enough arguments in call to smolder.NewErrorResponse
	have (number, error, string)
	want (int, bool, interface {}, string)
# github.com/muesli/beehive/api/resources/chains
api/resources/chains/chains_post.go:63:31: too many arguments in call to smolder.ErrorResponseHandler
	have (*restful.Request, *restful.Response, nil, *smolder.ErrorResponse)
	want (*restful.Request, *restful.Response, *smolder.ErrorResponse)
api/resources/chains/chains_post.go:63:80: not enough arguments in call to smolder.NewErrorResponse
	have (number, nil, string)
	want (int, bool, interface {}, string)
make: *** [build] Error 2
``` 

@muesli should probably have a closer look at the smolder changes, I'm not overly familiar with it.

prologic:
I really don't see the point of the TinderBee. It should just be removed and beehive resurrected as a useful alternative to the various SaaS based IFTT-type serivces.

bees/slackbee/slackbee.go
```diff
@@ -66,7 +66,7 @@ func (mod *SlackBee) Action(action bees.Action) []bees.Placeholder {
 		msgParams := slack.NewPostMessageParameters()
 		msgParams.LinkNames = 1
 		for _, to := range tos {
-			_, _, err := mod.client.PostMessage(to, text, msgParams)
+			_, _, err := mod.client.PostMessage(to, slack.MsgOptionText(text, false))
```
rubiojr: A better fix would be to set `msgParams := slack.PostMessageParameters{}`, then `_, _, err := mod.client.PostMessage(to, slack.MsgOptionPostMessageParameters(msgParams))`.

That shouldn't change the behavior.

bees/slackbee/slackbee.go
```diff
@@ -63,10 +63,10 @@ func (mod *SlackBee) Action(action bees.Action) []bees.Placeholder {
 			}
 		}
 
-		msgParams := slack.NewPostMessageParameters()
+		msgParams := slack.PostMessageParameters{}
 		msgParams.LinkNames = 1
 		for _, to := range tos {
-			_, _, err := mod.client.PostMessage(to, slack.MsgOptionText(text, false), slack.MsgOptionPostMessageParameters(msgParams))
+			_, _, err := mod.client.PostMessage(to, text, msgParams)
```
rubiojr: @cyian-1756 missed something in my last comment, this should be:

```go
_, _, err := mod.client.PostMessage(to, slack.MsgOptionText(text, false), slack.MsgOptionPostMessageParameters(msgParams))
```

rubiojr: Ah nevermind, I see you've addressed this in a different commit. Thanks!

muesli:
We could use the tinder library from beehive-vendor and push that to a new repository on GitHub. I'm not sure it's worth the effort though. An unmaintained library will break sooner than later and I don't think anyone here wants to really maintain a tinder library themselves.

muesli:
I appreciate all the effort you put into this! I'll probably squash and merge this tonight, but I also think at this point we should just move on to using go modules. Let me figure out in which order I want to approach things :laughing:

PR #215, opened 2019-04-14
Configuration API

rubiojr:
* Move json config related tasks to its own package (cfg)
* Provides an API to get/set the configuration (`/v1/config` endpoint)
* Saves the config every time we change it, instead of waiting for a graceful
  shutdown. This prevents losing config changes for unclean daemon shutdowns.

One of the current limitations of this approach is that it has been designed
with a single API client in mind, so saving the config isn't safe in a
multi-client environment.
If several clients try to update the config at the same time, only one will
succeed and changes from other clients will be discarded.

I'm not sure how big of a problem that'd be for beehive in reality.

### DIFF
--- a/api/resources/config/config.go
+++ b/api/resources/config/config.go
@@ -1,0 +1,64 @@
+/*
+ *    Copyright (C) 2018 Sergio Rubio
+ *
+ *    This program is free software: you can redistribute it and/or modify
+ *    it under the terms of the GNU Affero General Public License as published
+ *    by the Free Software Foundation, either version 3 of the License, or
+ *    (at your option) any later version.
+ *
+ *    This program is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *    Authors:
+ *      Sergio Rubio <sergio@rubio.im>
+ */
+
+package config
+
+import (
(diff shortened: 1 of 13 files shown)


rubiojr:
PoC from some old code I had around. Missing tests etc, but wanted to validate the idea with you first @muesli.

rubiojr:
`cfg.SaveCurrentConfig()` should probably be closer to the models in bees/chains/actions instead of polluting the api resources, adding some kind of observer pattern.

muesli:
Nice! I still see the regular REST-API as the common way of interacting with Beehive's config, especially for the web/admin interface. It is an easy way for exporting/importing entire configs however, which we'll certainly want to do there at some point!

You're right, we need to do something about the lazy config updating.

rubiojr:
@muesli what do you think if work on the API to import/export config via API here and tackle the lazy config updating in a different PR? Is 1. and 2. things we're interested in?

rubiojr:
Also thinking about moving config related stuff to its own `cfg` package in a different PR, if that sounds reasonable.

beehive.go
```diff
@@ -156,7 +121,7 @@ func main() {
 	config.Bees = bees.BeeConfigs()
 	config.Chains = bees.GetChains()
 	config.Actions = bees.GetActions()
-	saveConfig(config)
+	cfg.SaveConfig(config)
```
muesli: This may actually be obsolete then?

api/resources/config/config.go
```diff
@@ -0,0 +1,64 @@
+/*
```
muesli: I'd recommend moving the ConfigResource API endpoint to a separate PR, if you don't mind?

muesli:
Yeah, I agree. I'm certainly all for the suggested changes, but let's split this up into smaller PRs, which makes reviewing all the changes a bit easier to digest :laughing:

rubiojr:
Closing this in favor of https://github.com/muesli/beehive/pull/251

PR #216, opened 2019-04-15
Adds "Matches" support to filters

rubiojr:
So we can add things like:

`{{test Matches .text "^My.*foobar$"}}`

When adding filters to chains.


rubiojr:
Originally added in https://github.com/muesli/beehive/pull/61, looks like we missed this with the last filters refactor.

PR #219, opened 2019-04-25
Fix OSX Build error (goserial fails to compile and find termios.h)

prologic:
**Test Plan:**

```#!bash
prologic@Jamess-MacBook
Thu Apr 25 13:28:06
~/Contributions/beehive
 (fix_osx_build) 0
$ go build

prologic@Jamess-MacBook
Thu Apr 25 13:28:11
~/Contributions/beehive
 (fix_osx_build) 0
$ ./beehive -version
Beehive 0.0.0 (unknown)
```

This is likely a *temporary* work-around until a better solution is in place. Because `goserial` **should** compile. OS X **does** have a `termios.h` but I can't find the right solution to make `go build` find it.

### DIFF
--- a/bees/serialbee/serialbee.go
+++ b/bees/serialbee/serialbee.go
@@ -1,3 +1,8 @@
+// +build !darwin
+
+// FIXME: This bee doesn't build on macOS
+// XXX: https://gist.github.com/prologic/b5fa148410a26f917b4d944b72d847c0
+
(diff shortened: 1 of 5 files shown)


muesli:
This must be a problem on your machine, though. It builds fine on macOS here, as well as on Travis' macOS build systems.

PR #221, opened 2019-04-25
Dropped cgo dependency by switching to go-serial package

prologic:
Fix the OSX build issue with serialbee by using a better serial librrary that doesn't use CGO

Uses [go-serial](https://github.com/jacobsa/go-serial) instead

It builds :)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -39,7 +39,7 @@ require (
 	github.com/gorilla/websocket v1.2.0
 	github.com/guelfey/go.dbus v0.0.0-20131113121618-f6a3a2366cc3
 	github.com/huandu/facebook v2.1.2+incompatible
-	github.com/huin/goserial v0.0.0-20121012073615-7b90efdb22b1
+	github.com/jacobsa/go-serial v0.0.0-20180131005756-15cf729a72d4
 	github.com/jayeshsolanki93/devgorant v0.0.0-20160810172004-69fb03e5c3b1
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -84,3 +85,5 @@ require (
 	gopkg.in/fatih/set.v0 v0.1.0
 	layeh.com/gumble v0.0.0-20180224004502-127dae0a3703
 )
+
+replace github.com/huin/goserial => /Users/prologic/gohack/github.com/huin/goserial
```
muesli: Careful, this shouldn't be part of this commit.

go.sum
```diff
@@ -68,6 +69,8 @@ github.com/huandu/facebook v2.1.2+incompatible h1:APtODfdBm89WyxkAhP2rZlDvPFJv21
 github.com/huandu/facebook v2.1.2+incompatible/go.mod h1:wJogp9rhXUUjDuhx6ZaR5Eylx3dsJmy0zyFRaPYUq5g=
 github.com/huin/goserial v0.0.0-20121012073615-7b90efdb22b1 h1:v6symRpmVeKVUqq1grXnDDtus+lmQufUhgZ/lgMr4nU=
 github.com/huin/goserial v0.0.0-20121012073615-7b90efdb22b1/go.mod h1:x4wgpgRJT44loaDTf8/wWCkvTlhrKlVlaHATk7Leqlw=
```
muesli: The huin imports should be removed in that case.

PR #222, opened 2019-04-26
Add stdin as input-parameter to exec-bee

mkrauser:
The commands executed by execbee can now receive data via stdin

### DIFF
--- a/bees/execbee/execbee.go
+++ b/bees/execbee/execbee.go
@@ -66,9 +66,8 @@ func (mod *ExecBee) Action(action bees.Action) []bees.Placeholder {
 				}
 
 				go func() {
-					mod.Logf("StdIn" + stdin)
-					defer stdinPipe.Close()
 					io.WriteString(stdinPipe, stdin)
+					stdinPipe.Close()
 				}()
 			}
 
--- a/bees/execbee/execbee.go
+++ b/bees/execbee/execbee.go
@@ -18,6 +18,7 @@
  *    Authors:
  *      Dominik Schmidt <domme@tomahawk-player.org>
  *      Christian Muehlhaeuser <muesli@gmail.com>
+ *      Matthias Krauser <matthias@krauser.eu>
  */
 
 // Package execbee is a Bee that can launch external processes.
@@ -25,6 +26,7 @@ package execbee
 
 import (
 	"bufio"
+	"io"
 	"os/exec"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


bees/execbee/execbeefactory.go
```diff
@@ -103,6 +104,11 @@ func (factory *ExecBeeFactory) Actions() []bees.ActionDescriptor {
 					Description: "command to be executed",
 					Type:        "string",
 					Mandatory:   true,
+				}, {
```
muesli: Please make sure to always run gofmt on your sources before committing them. This will avoid formatting errors like this one.

bees/execbee/execbee.go
```diff
@@ -45,13 +47,31 @@ func (mod *ExecBee) Action(action bees.Action) []bees.Placeholder {
...
 
+			if stdin != "" {
+				stdinPipe, err := cmd.StdinPipe()
+
+				if err != nil {
+					mod.LogFatal("Error creating stdinPipe for Cmd", err)
+					return
+				}
+
+				go func() {
```
muesli: Just curious: is there a reason you execute this in the background?

bees/execbee/execbee.go
```diff
@@ -45,13 +47,31 @@ func (mod *ExecBee) Action(action bees.Action) []bees.Placeholder {
...
+					mod.Logf("StdIn" + stdin)
```
muesli: The supplied options / parameters are usually already printed out. I guess we could remove it here.

muesli:
Thanks for your contribution! Looking good, just a few remarks I added above. Simply force push to your branch, if you make any changes.

bees/execbee/execbeefactory.go
mkrauser: Sorry, somehow vscode used goreturns after an update

bees/execbee/execbee.go
mkrauser: No, I just took that from the go-exec documentation. But you are right, it it not necessary to run this in background.

bees/execbee/execbeefactory.go
muesli: goreturns should actually be fine, as I believe it also uses gofmt internally. Are you sure it caused that styling?

mkrauser: It seems there's more to it. I'll try to fix that later

bees/execbee/execbee.go
muesli: Ok, that got me intrigued. There must be a reason they use a separate goroutine in the example... and turns out there is:

The command isn't actually being started until you call CombinedOutput(). Writing to the StdinPipe would block hence.

So: your code is fine, ignore my remark!

bees/execbee/execbeefactory.go
mkrauser: It seems that gofmt accepts all those syntaxes:

`}, {`

```
},
{
```

```
},

{
```

at least on my machine, it only changed the indents, not the newlines. Only if there is more linebreaks than in the last case, those get removed.

muesli: Wow, you are right. Funny that I've noticed that ambiguous behavior before.

PR #223, opened 2019-04-27
Add tzdata to Docker image

jlelse:
When I developed something using alpine as base docker image, I wasn't able to set the correct time and timezone via the TZ environment variable until I added tzdata to the image.


muesli:
Thank you for contributing to Beehive!

PR #229, opened 2019-04-29
Cleans up unused deps

prologic:
Basically just ran:

```#!bash
$ go mod tidy
```


muesli:
Cleaned up the dependency issue with httplex and then (successfully) ran go mod tidy. Pushed to master.

Issue #144, opened 2017-04-17
Prometheus Hive


muesli:
A PrometheusHive could be used to let Beehive events expose data to Prometheus.

CalmBit:
Would you still be interested in having this happen? If you needed someone to work on it, I could give it a shot 😄

muesli:
Absolutely! Prometheus' Go API is rather straight forward and neat, too. Let me know if you run into any issues, I've worked with it before.

This could be really awesome to proxy/export all kinds of metrics data from services that don't natively support Prometheus.

Check out bees/hellobee for a simple "hello world" style bee.

CalmBit:
Alright, my workspace is set up and I'm starting work on it. I'll ping you if I have any questions!

muesli:
Merged to master. Thanks @CalmBit!

PR #231, opened 2019-04-30
Prometheus Bee

CalmBit:
A solution for issue #144 !

Might not merge immediately due to outstanding issues with the go.mod file, but it can at least be inspected :)


.gitignore
```diff
@@ -34,9 +34,11 @@ beehive.conf
 .*.sw*
 *.sublime-project
 *.sublime-workspace
+.vscode/
```
muesli: A bit of general advice right away: try to keep commits as granular/atomic as possible.

While I'm not generally opposed to adding the .vscode dir to the .gitignore, this isn't related to a Prometheus Bee and should be a separate commit & PR. I know this sounds nitpicky, but imagine for some terrible, imaginary reason we would have to revert your commit here: we'd also lose this unrelated change and then it becomes a big hassle to merge it back in.

bees/prometheusbee/prometheusbee.go
```diff
@@ -0,0 +1,253 @@
+/*
+ *    Copyright (C) 2019 	  CalmBit
```
muesli: Again, unbelievably nitpicky, but I'd try to keep the names left-aligned here.

bees/prometheusbee/prometheusbee.go
```diff
@@ -0,0 +1,253 @@
...
+		prometheus.CounterOpts{
+			Name: mod.counterVecName,
+			Help: "Collection of all counter variables being iterated by hives",
+		},
+		[]string{"name"},
+	)
+	err := prometheus.Register(mod.counter)
+	if err != nil {
+		mod.LogErrorf("Error registering counter vector: %v", err)
+		return
```
muesli: We should probably panic for these kind of errors, instead of simply returning. This is a pretty fatal situation, no?

bees/prometheusbee/prometheusbeefactory.go
```diff
@@ -0,0 +1,284 @@
+/*
+ *    Copyright (C) 2014-2017 Christian Muehlhaeuser
```
muesli: Forgot to add your copyright here! :smile:

muesli:
Looking good! Added a few super nitpicky remarks, but nothing major stood out! Obviously still need to sort out those go.mod issues, but that's not your fault.

bees/prometheusbee/prometheusbee.go
CalmBit: Not nitpicky! I just thought it'd be better right-aligned, but I'll fix it :)

.gitignore
CalmBit: Alright! That makes sense. Apologies for squeezing that in.

bees/prometheusbee/prometheusbee.go
CalmBit: Fair enough! I guess I mistakenly thought it wouldn't require a panic. Apologies :)

bees/prometheusbee/prometheusbeefactory.go
CalmBit: Whoops!

muesli:
I've fixed the httplex dependency issues in master. If you rebase and re-add the new required dependencies, we can merge this!

bees/prometheusbee/prometheusbeefactory.go
```diff
@@ -0,0 +1,286 @@
...
+					Description: "Value to subtract from the gauge",
+					Type:        "float64",
+					Mandatory:   true,
+				},
+			},
+		},
+		{
+			Namespace:   factory.Name(),
+			Name:        "gauge_set_to_current_time",
+			Description: "Sets a gauge's vaue to the current time as a unix timestamp",
```
muesli: Typo in "value".

PR #232, opened 2019-05-02
Fix handling of boolean config values

mkrauser:
Hey there,

I'm working on a bee right now (still WiP, but I will send a PR once it is finished) and discovered the following issue:
When a bee contains a boolean config value you cannot change the value from true to false without restarting beehive. It does work from false to true.

This PR fixes that.


muesli:
Thanks for your contribution!

Issue #237, opened 2019-05-03
bee.LogDebugF


CalmBit:
It'd be really cool to be able to put in some diagnostic information as debug-level logging, especially for situations in which someone is having an issue figuring out what's wrong with a bee. We've got a lot of debug information to report, but cluttering the log with this normally seems rather poor practice. So why not have a way to log debug statements?

muesli:
Yeah, I must admit I'm not sure why I didn't add that in the first place. Resolved by your own PR that I just merged into master!

PR #238, opened 2019-05-03
LogDebugF

CalmBit:
Adding a LogDebugF, along with some cleanup for logs.go removing some of the magic numbers :)

Aims to tackle #237 , although it also pertains to cleaning up the logging system a tiny bit. Completely optional PR!


muesli:
Much needed indeed!

PR #241, opened 2019-05-06
Email-Bee: Added missing event name

mkrauser:
Hey,

I forgot the name of the event in my last PR. Here's the fix for that 

/Matthias


muesli:
I apologize, I shouldn't have missed that in the review. This reminds me that we should have a thing to verify the validity/sanity of all defined events, actions & options.

mkrauser:
@muesli No problem, we are still all human ;-)

PR #242, opened 2019-05-06
Add JSON-Function for Templates

mkrauser:
This adds a json-Function to be used in String-Templates. In my Use-Case I serialize an entire Event and provide the Json-Document as Input to the ExecBee for further processing like this:

`{{.|Json}}`

I was not sure about the casing of "Json". I decided to go with that because it is easier to type and fits the existing functions. But technically it would be JSON.

### DIFF
--- a/templatehelper/templatehelper.go
+++ b/templatehelper/templatehelper.go
@@ -22,6 +22,8 @@
 package templatehelper
 
 import (
+	"encoding/json"
+	htmlTemplate "html/template"
 	"regexp"
 	"strings"
 	"text/template"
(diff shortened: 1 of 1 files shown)


muesli:
Neat! Makes me want to rethink the way we handle JSON in the HttpBee.

muesli:
I think `JSON` looks a little less weird. Want to change it to that?

mkrauser:
> I think `JSON` looks a little less weird. Want to change it to that?

Ok, no problem. I'm fine with both

PR #248, opened 2019-05-15
Youtube Bee (#25)

Mark-Jung:
Able to listen for Youtube notifications when beehive is exposed to a public url.

### DIFF
--- a/bees/youtubebee/youtube.go
+++ b/bees/youtubebee/youtube.go
@@ -1,0 +1,140 @@
+/*
+ *    Copyright (C) 2014      Daniel 'grindhold' Brendle
+ *                  2014-2017 Christian Muehlhaeuser
+ *
+ *	  This program is free software: you can redistribute it and/or modify
+ *	  it under the terms of the GNU Affero General Public License as published
+ *	  by the Free Software Foundation, either version 3 of the License, or
+ *	  (at your option) any later version.
+ *
+ *	  This program is distributed in the hope that it will be useful,
+ *	  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *	  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	See the
+ *	  GNU Affero General Public License for more details.
+ *
+ *	  You should have received a copy of the GNU Affero General Public License
+ *	  along with this program.	If not, see <http://www.gnu.org/licenses/>.
+ *
+ *	  Authors:
+ *		Daniel 'grindhold' Brendle <grindhold@skarphed.org>
+ *      Christian Muehlhaeuser <muesli@gmail.com>
+ */
+
+// Package youtubebee is a Bee for tunneling Youtube push notifications.
+package youtube
+
+import (
+	"encoding/xml"
+	"net"
+	"net/http"
+	"net/url"
+
+	"github.com/muesli/beehive/bees"
+)
+
+// YoutubeBee is a Bee for handling Youtube push notifications.
+type YoutubeBee struct {
+	bees.Bee
+
+	url string
+
+	addr string
+
+	eventChan chan bees.Event
+}
+
+// Run executes the Bee's event loop.
+func (mod *YoutubeBee) Run(eventChan chan bees.Event) {
+	mod.eventChan = eventChan
+	subscriptionLink := "https://pubsubhubbub.appspot.com/subscribe"
+	channelID := mod.url.split("/")
+	channelID = challenge[len(channelID)-1]
+	topic := "https://www.youtube.com/xml/feeds/videos.xml?channel_id=" + channelID
+
+	srv := &http.Server{Addr: mod.addr, Handler: mod}
+	l, err := net.Listen("tcp", mod.addr)
+	if err != nil {
+		mod.LogErrorf("Can't listen on %s", mod.addr)
+		return
+	}
+	defer l.Close()
+
+	go func() {
+		err := srv.Serve(l)
+		if err != nil {
+			mod.LogErrorf("Server error: %v", err)
+		}
+		// Go 1.8+: srv.Close()
+		// send POST to Google's pubsubhubbub to subscribe
+		resp, err := http.PostForm(subscriptionLink,
+			url.Values{
+				"hub.mode":     {"subscribe"},
+				"hub.topic":    {topic},
+				"hub.callback": {mod.addr},
+			})
+	}()
+
+	select {
+	case <-mod.SigChan:
+		return
+	}
+}
+
+func (mod *YoutubeBee) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method == "POST" {
+		ev := bees.Event{
+			Bee: mod.Name(),
+		}
+		ev.Name = "push"
+		type Feed struct {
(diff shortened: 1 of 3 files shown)


bees/youtubebee/youtube.go
```diff
@@ -0,0 +1,162 @@
...
+ *	  This program is distributed in the hope that it will be useful,
+ *	  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *	  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.	See the
+ *	  GNU Affero General Public License for more details.
+ *
+ *	  You should have received a copy of the GNU Affero General Public License
+ *	  along with this program.	If not, see <http://www.gnu.org/licenses/>.
+ *
+ *	  Authors:
+ *		Daniel 'grindhold' Brendle <grindhold@skarphed.org>
```
muesli: Add yourself here (and fix the indentation while you're at it :laughing:)

bees/youtubebee/youtube.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// Run executes the Bee's event loop.
+func (mod *YoutubeBee) Run(eventChan chan bees.Event) {
+	mod.eventChan = eventChan
+	subscriptionLink := "https://pubsubhubbub.appspot.com/subscribe"
+	channelURLTokens := strings.Split(mod.url, "/")
+	channelID := channelURLTokens[len(channelURLTokens)-1]
+	topic := "https://www.youtube.com/xml/feeds/videos.xml?channel_id=" + channelID
+	hardcodedAddress := "0.0.0.0:5050" // should be mod.addr
```
muesli: I agree with your comment, let's not hard-code that.

bees/youtubebee/youtube.go
```diff
@@ -0,0 +1,162 @@
...
+		data.Set("hub.topic", topic)
+		data.Set("hub.callback", mod.addr)
+		client := &http.Client{}
+		r, _ := http.NewRequest("POST", subscriptionLink, strings.NewReader(data.Encode())) // URL-encoded payload
+		r.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+		r.Header.Add("Content-Length", strconv.Itoa(len(data.Encode())))
+
+		resp, _ := client.Do(r)
+		for resp.Status != "202 Accepted" {
+			// redo until success
```
muesli: We need some proper error handling here, otherwise we may end up doing this infinitely.

bees/youtubebee/youtube.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func (mod *YoutubeBee) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	if req.Method == "POST" {
+		ev := bees.Event{
+			Bee: mod.Name(),
+		}
+		ev.Name = "push"
+		type Feed struct {
```
muesli: This struct is so long, I'd move it out of the function.

bees/youtubebee/youtubefactory.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+	return opts
+}
+
+// Events describes the available events provided by this Bee.
+func (factory *YoutubeBeeFactory) Events() []bees.EventDescriptor {
+	events := []bees.EventDescriptor{
+		{
+			Namespace:   factory.Name(),
+			Name:        "push",
```
muesli: I feel that the name `push` is giving the wrong impression. It sounds like an action we're triggering on YouTube. Is there more data we can extract from the subscription notification? What are typical events you would receive here? Info about new comments / subscriptions / likes?

Mark-Jung: The information that would be received are whether the channel posted a new video, updated a video title, or updated a video description. Basically, we are listening to the push notifications of the channel.

Mark-Jung:
I incorporated the comments. Please let me know if I should fix anything else! @muesli

bees/youtubebee/youtubefactory.go
muesli: I think we should probably parse this information and offer events like `new_video`, `changed_title` & `change_description` then.

muesli: Otherwise this will just become a parsing exercise for all the bee's users...

bees/youtubebee/youtubefactory.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// Events describes the available events provided by this Bee.
+func (factory *YoutubeBeeFactory) Events() []bees.EventDescriptor {
+	events := []bees.EventDescriptor{
+		{
+			Namespace:   factory.Name(),
+			Name:        "notification",
+			Description: "A push notification was sent by the Youtube channel",
+			Options: []bees.PlaceholderDescriptor{
```
CalmBit: Is there any other information that can be extracted from YouTube's API? As it stands, this mostly just seems like a generic push, rather than something more actionable. If not, that's absolutely fine - I'm just wondering, so we can broaden it now if possible 😄

CalmBit: Also, again, if we _can_ extract more information, then @muesli's suggestion of dividing the current monolith event into multiple events might be a good idea.

bees/youtubebee/youtubefactory.go
```diff
@@ -0,0 +1,113 @@
...
+ *    This program is distributed in the hope that it will be useful,
+ *    but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *    GNU Affero General Public License for more details.
+ *
+ *    You should have received a copy of the GNU Affero General Public License
+ *    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *    Authors:
+ *      Christian Muehlhaeuser <muesli@gmail.com>
```
CalmBit: You should add yourself here too! 😄

bees/youtubebee/youtubefactory.go
Mark-Jung: I agree how the current monolith event should be split into multiple events so I looked at the notification and the response object provided in the [documentation](https://developers.google.com/youtube/v3/guides/push_notifications). Unfortunately, I don't think it's possible to divide the notification into three different types of events. I think you can definitely expand this bee for more functionalities such as uploading video, but in the aspect of notification this is as specific as it can get I believe.

bees/youtubebee/youtubefactory.go
Mark-Jung: haha yup.

bees/youtubebee/youtubefactory.go
CalmBit: I'm not sure if the publication/updated date are initially set to the same value for initial publication, but if they were you could check if they were the same, and at least divide them into two - the initial publish and the update.

Mark-Jung: oof good catch. I missed that. I'll commit that this weekend!

Mark-Jung:
@CalmBit @muesli I specified the monolithic event into two: new_vid and change_vid. Please let me know if you see anything else I need to fix!

PR #257, opened 2019-06-18
Fix compilation

Invertisment:
Travis builds had updated their lib and they use more pointer-based variables.


muesli:
I wouldn't mind updating to the newer go-travis API, however it seems you deleted your branch? You'd have to bump `go.mod` & `go.sum` for this to compile properly, though.

PR #261, opened 2019-08-12
Add Rocket.Chat hive

dsbrng25b:
Add hive which allows to send messages into a Rocket.Chat channel.


bees/rocketchatbee/client.go
```diff
@@ -0,0 +1,190 @@
+/*
+ *    Copyright (C) 2016 Sergio Rubio
+ *                  2017 Christian Muehlhaeuser
```
muesli: Feel free to add yourself here and a couple lines below under `Authors:`.

bees/rocketchatbee/rocketchatbeefactory.go
```diff
@@ -0,0 +1,127 @@
...
+				},
+				{
+					Name:        "text",
+					Description: "Content of the message",
+					Type:        "string",
+					Mandatory:   true,
+				},
+				{
+					Name:        "alias",
+					Description: "The name to show as sender. If empty the to which the user_id belongs to.",
```
muesli: Second sentence sounds a bit broken. It probably means `If empty, the name to which the user_id belongs to`?

muesli:
Nice work! A few more comments:

- `client.go` could almost become its own library. I'm fine with it for now, seeing how the scope is still fairly limited, but it's something one should eventually think about.

- it would be nice to have events, too!

- the logo seems to be cut off at the right border?

dsbrng25b:
Thank you for revewing the code. I added my self as author and changed the description of the alias setting. 
                              
When I find some time, I may implement support for events as well. But right now, for me sending messages has priority.  
I agree that if the Rocket.Chat API client grows it should become its own library, but for now I think its ok like that. 
If I implement the event part I probably use https://github.com/detached/gorocket, but currently two things in gorocket are missing:         
* authentication with user-id and auth-token                                    
* get the channel ID by the channel name.

The logo looks good on me. Did you look at the logo in Github and didn't see the scrollbar under the picture?

muesli:
Hah, sorry about that, you're right about the logo! LGTM!

PR #265, opened 2019-10-18
Adds ipify bee

rubiojr:
Queries ipify.org for the public IP and sends an event if the IP changed.


muesli:
Thank you, another awesome bee! :smiley:

PR #267, opened 2019-10-18
pushover: do not print token to stdout

rubiojr:
`mod.Logln(msg)` will write the Pushover tokens to stdout, which isn't ideal
if the logs are saved somewhere (when running from systemd for example).

Send the debug output to the debug log instead.

### DIFF
--- a/bees/pushoverbee/pushoverbee.go
+++ b/bees/pushoverbee/pushoverbee.go
@@ -71,14 +71,14 @@ func (mod *PushoverBee) Action(action bees.Action) []bees.Placeholder {
 			msg.Set("title", title)
 		}
 
-		mod.Logln(msg)
+		mod.LogDebugf("Message: %s", msg)
 		resp, err := http.PostForm("https://api.pushover.net/1/messages.json", msg)
 		if err != nil {
 			panic(err)
 		}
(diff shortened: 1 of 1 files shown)


bees/pushoverbee/pushoverbee.go
```diff
@@ -71,14 +71,14 @@ func (mod *PushoverBee) Action(action bees.Action) []bees.Placeholder {
...
-		mod.Logln(msg)
+		mod.LogDebugf("Message: %s", msg)
 		resp, err := http.PostForm("https://api.pushover.net/1/messages.json", msg)
 		if err != nil {
 			panic(err)
 		}
 		defer resp.Body.Close()
 		if resp.StatusCode == 200 {
-			mod.Logln("Pushover send message success.")
+			mod.LogDebugf("Pushover send message success.", nil)
```
muesli: This leads to a warning with Go >=1.12:
```
bees/pushoverbee/pushoverbee.go:81:17: LogDebugf call has arguments but no formatting directives
```

We can simply remove the "nil" argument here.

rubiojr: Done!

PR #271, opened 2019-10-19
ipify and cfddns: fix directory name

rubiojr:
Follow the existing convention ending the bee dir name with a 'bee' suffix.


rubiojr:
Took the chance to fix the ordering in `hives.go` that I accidentally broke.

PR #272, opened 2019-10-19
Add Redis hive

rubiojr:
Key/value storage and pubsub support using a Redis server.


muesli:
We could hook this up to Redis pub/sub system. That would actually be incredibly powerful and awesome...

rubiojr:
Agreed :smile:.

Added very basic subscribe support in 25c3e4b3f0d30e7db3fa30d789b4271bbbd017a9 to deliver events received from a Redis pubsub channel and it opens the door to very interesting possibilities.

Still a WiP, need to make all this a bit more resilient and also thinking about adding publish support before merging this also, so other bees can send bits to the Redis channel.

muesli:
@rubiojr Awesome work! Just let me know whenever you consider this ready to be merged.

rubiojr:
I'm almost there. Some minor refactoring and doc updates and should be good for :eyes:

rubiojr:
This is ready for :eyes: 

To real world test this, I'm currently writing a small tool that:

* Posts a URL to Beehive's http server hive
* Stores the URL in Redis
* Publishes the URL to a redis-channel
* Consumes the URL from Redis and downloads the video with youtobe-dl if it's a YouTube URL
* Notifies via pushover that a URL was received and a video will be downloaded if the URL is from YouTube
* Publishes a message to a Redis channel when the download has finished so Beehive sends a pushover notification

rubiojr:
Added sample script and Beehive config to https://github.com/rubiojr/beehive-youtube-dl

rubiojr:
go-redis built-in retries/backoff works quite well so switched to that instead of using an external backoff library, which simplified the code a bit.

rubiojr:
Just realized that my last force push to incorporate master changes didn't work as intended... fixing in a bit.

rubiojr:
> Just realized that my last force push to incorporate master changes didn't work as intended... fixing in a bit.

Fixed in https://github.com/muesli/beehive/pull/272/commits/351b85af354892d24e14bd88310e9eacdab9fe0d

bees/redisbee/redisbee.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+	options.Bind("password", &password)
+	var db int
+	options.Bind("db", &db)
+
+	client := redis.NewClient(&redis.Options{
+		Addr:            fmt.Sprintf("%s:%s", host, port),
+		Password:        password, // no password set
+		DB:              db,       // use default DB
+		MinRetryBackoff: 80 * time.Millisecond,
```
muesli, line 127: We should probably declare those values as consts, with a little comment why we picked the specific value we did.

rubiojr: Good call, fixed!

rubiojr:
Thanks y'all for the :eyes: !

PR #273, opened 2019-11-02
Systemd's Watchdog bee

rubiojr:
Beehive integration with Systemd's watchdog.

Linux only.

http://0pointer.de/blog/projects/watchdog.html


bees/watchdogbee/watchdogbee.go
```diff
@@ -0,0 +1,71 @@
...
+	for {
+		select {
+		case <-mod.SigChan:
+			return
+		case <-time.After(time.Duration(interval) * time.Second):
+			resp, err := http.Get(api.CanonicalURL().String())
+			if err == nil {
+				resp.Body.Close()
+				mod.LogDebugf("Notify Systemd's watchdog")
+				daemon.SdNotify(false, "WATCHDOG=1")
```
muesli: Coincidentally I fixed upstream `go-systemd` a year ago to not rely on those magic-strings anymore:

https://github.com/coreos/go-systemd/commit/39ad653e9c19ed64c183ee84c876e5ff37ad7a49#diff-1f15fcf5421eecd6ca1b61337c319495

So we can simple use:

```
daemon.SdNotify(false, SdNotifyWatchdog)
```

muesli:
Thinking about this a bit more, maybe this should rather be core functionality instead of a bee?

rubiojr:
> Thinking about this a bit more, maybe this should rather be core functionality instead of a bee?

Yeah, I thought about it, but given that multi OS support for this would be challenging I opted for this approach.

If you'd like to see this integrated into the core, conditionally enabling it when systemd is detected, I'd be happy to give it a try, it'd definitely make automatic configuration easier.

muesli:
We could solve that in a similar fashion as we only include some hives on particular platforms (`hives_linux.go` etc). I think it makes sense to move this to core, but maybe a systemd bee could still be cool: maybe we can trigger events when services are changing state, like (re)starting or stopping?

We could theoretically also start / stop services, even though we'd usually be limited to interacting with user services (we're not running as `root` after all).

rubiojr:
Good ideas. I’ll start by abstracting the watchdog so we can add it to the core.

rubiojr:
cc https://github.com/muesli/beehive/pull/275

rubiojr:
https://github.com/muesli/beehive/pull/275 was merged so closing this.

PR #275, opened 2019-11-08
Systemd's Watchdog

rubiojr:
Linux only.

Enable Systemd's watchdog integration when running under Systemd and
and the service unit has enabled the watchdog feature.

Related: https://github.com/muesli/beehive/pull/273

### DIFF
--- a/watchdog_linux.go
+++ b/watchdog_linux.go
@@ -1,0 +1,44 @@
+package main
+
+import (
+	"net/http"
+	"time"
+
+	"github.com/coreos/go-systemd/daemon"
+	"github.com/muesli/beehive/api"
+	log "github.com/sirupsen/logrus"
+)
+
+// Notifies Systemd's watchdog every WatchdogSec/3 when running
+// under Systemd and the watchdog feature has been enabled in
+// the service unit.
+//
+// This will no-op when not running under Systemd.
+//
+// See http://0pointer.de/blog/projects/watchdog.html
+// and https://www.freedesktop.org/software/systemd/man/systemd.service.html
+//
+func init() {
+	interval, err := daemon.SdWatchdogEnabled(false)
+	// systemd's service unit interval in microseconds
+	if err != nil || interval == 0 {
+		log.Printf("Systemd watchdog not enabled")
+		return
+	}
+	interval = interval / 3 / 1000000000
+	log.Printf("Systemd watchdog interval: %d", interval)
+
+	go func() {
+		for {
+			select {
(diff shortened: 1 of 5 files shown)


watchdog_linux.go
```diff
@@ -0,0 +1,44 @@
...
+// and https://www.freedesktop.org/software/systemd/man/systemd.service.html
+//
+func init() {
+	interval, err := daemon.SdWatchdogEnabled(false)
+	// systemd's service unit interval in microseconds
+	if err != nil || interval == 0 {
+		log.Printf("Systemd watchdog not enabled")
+		return
+	}
+	interval = interval / 3 / 1000000000
```
muesli: I'm not quite sure I understand that line. Between what units are we converting here?

muesli:
Nice! We also got a sample service file in https://github.com/muesli/beehive/tree/master/admin which we should update accordingly.

watchdog_linux.go
rubiojr: Interval is microseconds according to the docs, while WatchdogSec is is seconds, so a  value of 30 secs in the systemd unit would be 30000000/3/1000000, or notify every 10 secs. Does that make sense? Sound like I should add some more comments.

rubiojr: Now that you mention it, I’ve added some extras that should be there. Let me have a closer look.

muesli: I see. We could probably use Go's `time.Millisecond` / `time.Second` to make it a bit more readable.

rubiojr: The math was correct buy I made it overly  confusing because I still do not understand how Golang's time.Duration works.

rubiojr:
@muesli pushed some fixes updating docs and simplifying the math. Let me know if that reads a bit better.

rubiojr:
Looks something like:

```
Nov 10 16:32:52 dunraid beehive[38759]: DEBU[0010] Systemd watchdog notify
Nov 10 16:33:02 dunraid beehive[38759]: INFO[0020] Handling request                              Method=GET URL=/
Nov 10 16:33:02 dunraid beehive[38759]: INFO[0020] serving config/index.html ... (from )
Nov 10 16:33:02 dunraid beehive[38759]: INFO[0020] Finished request                              Duration="274.426µs" Method=GET Response=200 URL=/
Nov 10 16:33:02 dunraid beehive[38759]: DEBU[0020] Systemd watchdog notify
```

when the debug log is enabled.

muesli:
Awesome, thank you for another great contribution! :heart:

PR #279, opened 2019-12-13
Add support for KnowGo Adaptation Layer events

pmundt:
This is an initial attempt at getting notifications from our backend into the system, which we plan to expand with additional events and actions over time.


bees/knowgobee/knowgobeefactory.go
```diff
@@ -0,0 +1,131 @@
...
+ *      Paul Mundt <paul.mundt@adaptant.io>
+ */
+
+package knowgobee
+
+import (
+	"github.com/muesli/beehive/bees"
+)
+
+// HelloBeeFactory is a factory for HelloBees.
```
CalmBit: Just a nitpick, this should be updated to KnowGoBeeFactory and KnowGoBees 😄

pmundt: Thanks, I thought I caught all of these but I guess not. This is what I get for eyeballing it and doing it manually..

bees/knowgobee/knowgobee.go
```diff
@@ -0,0 +1,166 @@
...
+
+	sub, err := client.Subscribe(&api.SubscriptionRequest{
+		Event: "country-change",
+	})
+	if err != nil {
+		mod.LogErrorf("Failed to subscribe:", err)
+		return
+	}
+
+	mod.Logf("Successfully subscribed")
```
rubiojr: I'd suggest to use `log.Debugln` here if this is only useful while debugging. See https://github.com/muesli/beehive/pull/266

muesli:
Looks good to me, too, apart from @rubiojr's comment. Let's try not to spam the console too much. In the meantime I'm working on a logger module that lets you browse/filter by each bee's individual output. That should improve the situation quite a bit.

bees/knowgobee/knowgobee.go
pmundt: Fine with me

pmundt:
I've now switched it over to the debug logger, as requested.

muesli:
One issue we got here is depending on `http.NewRequestWithContext` which is Go 1.13+.

Issue #188, opened 2018-04-12
Support HTTP headers in HTTPBee


leopku:
Many case of http request need special headers.
Is it possible to add support for headers for httpbee?
Thanks a lot.

muesli:
Agreed, we should add that.

PR #282, opened 2020-02-05
Add HTTP Headers

CalmBit:
FINALLY tackling #188 , after so long...

This just adds HTTP header support to the HTTP Client hive, allowing for HTTP headers to be attached to actions, and returns both request and response headers as event parameters upon completion.

### DIFF
--- a/bees/httpbee/httpbee.go
+++ b/bees/httpbee/httpbee.go
@@ -16,6 +16,7 @@
  *
  *    Authors:
  *      Christian Muehlhaeuser <muesli@gmail.com>
+ *		CalmBit <calmbit@posteto.net>
  */
 
 // Package httpbee is a Bee that lets you trigger HTTP requests.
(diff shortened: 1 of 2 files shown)


bees/httpbee/httpbee.go
```diff
@@ -86,16 +119,39 @@ func (mod *HTTPBee) Action(action bees.Action) []bees.Placeholder {
...
+		}
+
+		req, err := http.NewRequest("POST", u, buf)
+		req.Header.Set("Content-Type", ctype)
+
+		if err != nil {
+			mod.LogErrorf("Error: %s", err)
+			return outs
+		}
+		for _, header := range h {
```
muesli: I feel like there's a bit of code duplication going on here. We could probably move the header-parsing into a separate little helper method.

CalmBit: Whoops! Some of the duplication here is a bit obvious, huh...

Should all be fixed with this next commit 😄

PR #284, opened 2020-02-20
Create bee Events with less Effort

penguwin:
This adds a new `createEvent()` method on the implemented Bees which handles back a filled `Event`-Struct for the given Eventname and map of Placholder values. 

The map for the Placeholders should be structured like this:
```
properties := map[string]interface{}{
          "event0": val0,
          "event1": val1,
          "event2": val2,
          // ...
 }
```

With a map like this you can simple trigger an event like so:
```
func (mod *MyBee) TriggerEvent(name string, properties map[string]interface{}) {
    mod.evchan <- mod.CreateEvent(name, properties)
}
```

I also demonstrated the usage in `openweathermapbee/event.go` and in the `ircbee.go` and as you can see this reduces the amount of code we have to write.

### DIFF
--- a/bees/bees.go
+++ b/bees/bees.go
@@ -16,6 +16,7 @@
  *
  *    Authors:
  *      Christian Muehlhaeuser <muesli@gmail.com>
+ *      Nicolas Martin <penguwin@penguwin.eu>
  */
 
 // Package bees is Beehive's central module system.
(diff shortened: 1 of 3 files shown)


penguwin:
force pushed some fixed typos in the my comments

rubiojr:
It's a matter of preference, but I'm ok with the existing code if that helps. It's explicit, making it easier to understand what kind of types are being sent in each event when reading the code. A bit less magical.

penguwin:
I agree that the existing code is less magical and also more explicit about the types. Yet the `EventDescriptor` already contains the type-information for each placeholder which makes it a bit redundant to have to also declare it for every placeholder in the `Event` itself. 

On second thought, I realized that not using the code from this pr would be in favour of the KISS principle - there should be only one consistent way to declare the events (even if it's more verbose). Therefore I'd also say that we should rather not merge this pr. Any final thoughts @muesli?

---

p.s. after looking at the code again I also noticed that I should probably break the loop in `CreateEvent` after appending each Placeholder to the Event - and also return an error if no placeholders were created instead of just returning an empty event.

PR #285, opened 2020-02-21
Refactored Mastodonbee a bit

### LINKED ISSUE #247: Delete action for Mastodon
Add delete as an action for Mastodon bees.

A potential use case: Setting up a chain to delete toots/tweets after *x* number of days.

Same goes for any social media platform that allows deleting via API, really. E.g. Twitter, Facebook, etc.

penguwin:
I refactored the MastodonBee a bit. 
-  renamed `mastodonBee/Factory` to `MastodonBee/Factory` as starting with an uppercase letter is more consistent with the naming scheme of other hives/bees.
- Add timestamps in the `toot_fetched`-Event. This resolves #247 as we can already delete toots by ID.

I also created a [gist](https://gist.github.com/penguwin/67a5ee185458eca9772fc85649b34e04) with a skeleton for Mastodon&IRC interaction. You'd just need to fill in the credentials and the desired IRC channel.

### DIFF
--- a/bees/mastodonbee/events.go
+++ b/bees/mastodonbee/events.go
@@ -70,6 +70,11 @@ func (mod *MastodonBee) handleStatus(status *mastodon.Status) {
 				Value: status.URL,
 				Type:  "string",
 			},
+			{
+				Name:  "created",
+				Value: status.CreatedAt,
+				Type:  "time.Time",
+			},
 		},
 	}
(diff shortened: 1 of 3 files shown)


penguwin:
I've removed the usage of `createEvent()` (#284) from this pr - therefore this is ready for review now.

PR #286, opened 2020-03-19
Load and write config from/to standard paths

### LINKED ISSUE #283: Proposal: add user configuration directory to the config search path
Beehive tries to load a `beehive.conf` from the current directory `./beehive.conf` when booting (if found). If not found, it'll use an in-memory representation until the process exits cleanly, then write it to the current directory.

The proposal is to also search for a config using `$os.UserConfigDir/beehive/beehive.conf`, so `~/.config/beehive/beehive.conf` under Linux for example (platform dependent).

With this change, Beehive would search for the config file in this particular order:
1. The current working directory (i.e. the way it works now)
2. `$os.UserConfigDir/beehive/beehive.conf`

The change would be backwards compatible.

I also thought about writing to `$os.UserConfigDir/beehive` instead of `$CWD` when there's no config present, which sounds harmless for the most part, but that'd change behavior in a backwards incompatible way.

Thoughts?

rubiojr:
This changes Beehive's behavior.

When writting the config file for the first time
(i.e. if beehive was run without a config), the file will be written to
`~/.config/beehive/beehive.conf` (OS dependant) instead of the current
directory.

Loading the config from the current directory, if available, is still
preferred, so that doesn't change.

Fixes #283

### DIFF
--- a/beehive.go
+++ b/beehive.go
@@ -91,7 +91,7 @@ func main() {
 	var config cfg.Config
 	var err error
 	if configFile != cfg.DefaultPath() {
-		_, err = cfg.LoadConfig(configFile)
+		config, err = cfg.LoadConfig(configFile)
 		if err != nil {
 			log.Fatalf("Error loading specified config file %s. err: %v", configFile, err)
 		}
--- a/beehive.go
+++ b/beehive.go
@@ -50,9 +50,10 @@ var (
 func main() {
 	app.AddFlags([]app.CliFlag{
 		{
-			V:    &configFile,
-			Name: "config",
-			Desc: "Config-file to use",
+			V:     &configFile,
+			Name:  "config",
+			Value: cfg.DefaultPath(),
+			Desc:  "Config-file to use",
 		},
 		{
 			V:     &versionFlag,
@@ -87,25 +88,18 @@ func main() {
 	log.Println()
 	log.Println("Beehive is buzzing...")
 
-	config := cfg.Config{}
+	var config cfg.Config
 	var err error
-	if configFile == "" {
-		configFile = cfg.FindUserConfigPath()
-		if configFile != "" {
-			config, err = cfg.LoadConfig(configFile)
+	if configFile != cfg.DefaultPath() {
+		_, err = cfg.LoadConfig(configFile)
+		if err != nil {
+			log.Fatalf("Error loading specified config file %s. err: %v", configFile, err)
 		}
 	} else {
-		config, err = cfg.LoadConfig(configFile)
-	}
-
-	if err != nil {
-		log.Panicf("Error loading config file %s!: %v", configFile, err)
-	}
-	if configFile != "" {
-		log.Printf("Config file loaded from %s\n", configFile)
-	} else {
-		configFile = cfg.DefaultPath()
-		log.Println("No config file found, loading defaults")
+		configFile, config, err = cfg.Load()
+		if err != nil {
+			log.Fatalf("Error loading user config file %s. err: %v", configFile, err)
+		}
 	}
 
 	// Load actions from config
--- a/beehive.go
+++ b/beehive.go
@@ -90,15 +90,23 @@ func main() {
 
 	var config cfg.Config
 	var err error
-	if configFile != cfg.DefaultPath() {
-		config, err = cfg.LoadConfig(configFile)
+	if configFile != cfg.DefaultPath() { // the user specified a custom config path
+		config, err = cfg.Load(configFile)
 		if err != nil {
 			log.Fatalf("Error loading specified config file %s. err: %v", configFile, err)
 		}
-	} else {
-		configFile, config, err = cfg.Load()
-		if err != nil {
-			log.Fatalf("Error loading user config file %s. err: %v", configFile, err)
+	} else { // try to load config from user paths
+		path := cfg.Lookup()
+		if path == "" {
+			log.Info("No config file found, loading defaults")
+			config = cfg.Config{}
+		} else {
+			configFile = path
+			log.Infof("Loading config file from %s", path)
+			config, err = cfg.Load(path)
+			if err != nil {
+				log.Fatalf("Error loading user config file %s. err: %v", configFile, err)
+			}
 		}
 	}
 
@@ -119,7 +127,7 @@ func main() {
 		abort := false
 		switch s {
 		case syscall.SIGHUP:
-			config, err := cfg.LoadConfig(configFile)
+			config, err := cfg.Load(configFile)
 			if err != nil {
 				log.Panicf("Error loading config from %s: %v", configFile, err)
 			}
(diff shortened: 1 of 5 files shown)


rubiojr:
I'll need to fix the tests. I added `os.UserConfigDir()` which was introduced in Go `1.13`

beehive.go
```diff
@@ -50,10 +50,9 @@ var (
 func main() {
 	app.AddFlags([]app.CliFlag{
 		{
-			V:     &configFile,
-			Name:  "config",
-			Value: "./beehive.conf",
```
muesli: The nice thing about keeping the default value defined here, is that it actually shows up as a helpful little reminder when running `beehive --help`:

```
  -config string
        Config-file to use (default "./beehive.conf")
```

beehive.go
```diff
@@ -90,12 +89,23 @@ func main() {
 
 	config := cfg.Config{}
 	var err error
-	if cfg.Exist(configFile) {
-		config, err = cfg.LoadConfig(configFile)
-		if err != nil {
-			log.Panicf("Error loading config file %s!: %v", configFile, err)
+	if configFile == "" {
```
muesli: Related to the previous comment (I totally understand why you removed the default): maybe we can check whether the flag has been "touched" (as in passed in by the command)?

If it's untouched we can proceed trying to find the config file here.

Maybe I'm overthinking this though, and your change is just fine.

cfg/config.go
```diff
@@ -4,27 +4,88 @@ import (
...
+	userScope := gap.NewScope(gap.User, appName)
+	path, err := userScope.ConfigPath(cfgFileName)
+	if err != nil {
+		return cfgFileName
+	}
+
+	return path
+}
+
+func defaultSearchPaths() []string {
```
muesli: You could use `gap`'s `LookupConfig` function for this. It returns all existing config files in the default (XDG) search paths.

beehive.go
rubiojr: So the issue I had with this, is that with the default changed to OS specific locations, that help in string is a bit longer and not very pretty. It's only a cosmetic/readability issue though, I'm happy to set the default value to the new/preferred OS dependent location. Whatcha think?

cfg/config.go
rubiojr: Totally missed that, will have a look.

rubiojr: So using this as-is would introduce an additional complication, given that it'll also return system config paths even when using the user scope, if the file exists in `/etc/beehive/beehive.conf` for example.

If beehive is running as a regular user without write access to that `/etc` config file, we'd need to either:

* Panic and tell the user we can't write to the config file
* Teach beehive to gracefully degrade and operate in config read-only mode, including API and/or Web UI changes
* Write changes to the user config dir after reading the config from the system path, which could be a bit confusing for existing users perhaps.
* Remove system scope paths from the returned array, so only user paths `./beehive.conf` and `~/.config/beehive/beehive.conf` are returned.

I'm opting for the last one to keep things simple, but I don't have a strong preference here.

rubiojr:
@muesli I think the last commit addresses your feedback, simplifying the logic a bit also.

```
Usage of ./beehive:
  -bind string
    	Which address to bind Beehive's API & admin interface to (default "localhost:8181")
  -canonicalurl string
    	Canonical URL for the API & admin interface (default "http://localhost:8181")
  -config string
    	Config-file to use (default "/home/rubiojr/.config/beehive/beehive.conf")
  -debug
    	Turn on debugging
  -version
    	Beehive version
```

I tried using `gap`'s `LookupConfig`, but complicated the code to exclude system paths unfortunately, so defaulted to something simpler. Operating in ready-only mode or loading from system path and writing to user dir is probably a good topic for a new issue/discussion/PR.

rubiojr:
I don't have a Windows host around, but seems to be working fine on macOS:

```
➜  beehive git:(config-dir) ./beehive --help
Usage of ./beehive:
  -bind string
    	Which address to bind Beehive's API & admin interface to (default "localhost:8181")
  -canonicalurl string
    	Canonical URL for the API & admin interface (default "http://localhost:8181")
  -config string
    	Config-file to use (default "/Users/rubiojr/Library/Preferences/beehive/beehive.conf")
  -debug
    	Turn on debugging
  -version
    	Beehive version
➜  beehive git:(config-dir) ./beehive 
INFO[0000] Registering Resource                          Resource=HiveResource
INFO[0000] Registering Resource                          Resource=BeeResource
INFO[0000] Registering Resource                          Resource=ChainResource
INFO[0000] Registering Resource                          Resource=ActionResource
INFO[0000] Registering Resource                          Resource=LogResource
INFO[0000]                                              
INFO[0000] Beehive is buzzing...                        
INFO[0000] No config file found, loading defaults       
^CINFO[0000] Got signal: interrupt                        
INFO[0000] Storing config to /Users/rubiojr/Library/Preferences/beehive/beehive.conf 
```

beehive.go
rubiojr: Done :heavy_check_mark:

beehive.go
rubiojr: Done :heavy_check_mark:, doesn't look that bad.

cfg/config.go
```diff
@@ -4,27 +4,111 @@ import (
...
+	}
+	return paths[0]
+}
+
+// Load tries to load the default config file, returning an empty config object
+// if not found
+//
+// Returns the path of the config found and the Config object, returning an error
+// also if loading the config from the filesystem fails.
+func Load() (string, Config, error) {
```
muesli, line 92: I think the naming is a bit confusing here. Maybe `Load` should be called `LoadDefaultConfig` to differentiate it from `LoadConfig` a bit more. Part of me thinks there should only be one public method responsible for loading the config. Maybe the caller should retrieve the config path first and then only call `LoadConfig`?

This is probably awfully nitpicky.

cfg/config.go
```diff
@@ -4,27 +4,111 @@ import (
...
 
 	"github.com/muesli/beehive/bees"
+	gap "github.com/muesli/go-app-paths"
+	log "github.com/sirupsen/logrus"
 )
 
+const appName = "beehive"
+
+var cfgFileName = "beehive.conf"
+var lookupFunc = Lookup
```
muesli, line 17: On the one hand this is great because it allows for better tests, but on the other hand I'm a bit wary because it adds another layer of abstraction to the code here.

rubiojr: Makes sense. An integration test that doesn't require source code changes would probably be a better places for something like this.

cfg/config.go
rubiojr: Make sense to me too. I'm getting rid of `LoadConfig` and replacing it with a single `cfg.Load` public function. The `Config` suffix was also a bit repetitive given the package is also named `cfg`.

rubiojr:
@muesli the last commit should address your feedback.

Took the change to make the `cfg` package public interface a bit more idiomatic, renaming the `LoadConfig` and `SaveConfig` public functions, at the expense of breaking backwards compat (assuming no one cares about Beehive's `cfg` package that should probably be in `internal` or something).

rubiojr:
Superseded by https://github.com/muesli/beehive/pull/292

PR #292, opened 2020-04-12
Improve the configuration system

### LINKED ISSUE #283: Proposal: add user configuration directory to the config search path

### LINKED ISSUE #291: Proposal: abstract interface for loading/saving configuration
The intention is to abstract the configuration interface in `cfg` so we can implement multiple configuration backends, including an encrypted backend that would address https://github.com/muesli/beehive/issues/226.
As a secondary objective, I'd like to experiment with loading beehive's config from a networked endpoint, like an S3 bucket, so this abstraction should make that implementation cleaner.

This would build on top of https://github.com/muesli/beehive/pull/286 which exposes a smaller and cleaner public interface, with `Lookup`, `Save` and `Load` as the only public functions to be abstracted.

rubiojr:
## New config path defaults

Builds on top and supersedes https://github.com/muesli/beehive/pull/286

Load and write config from/to standard paths, changing Beehive's default behaviour.

This is a **breaking change**: when writing the config file for the first time (i.e. if beehive was run without a config), the file will be written to `~/.config/beehive/beehive.conf` (OS dependant) instead of the current directory.

Loading the config from the current directory, if available, is still preferred, so that doesn't change.
    
Fixes #283

## New config backend abstraction

This PR also adds support for multiple config file backends to make it easier to implement other backends (encrypted, networked, etc).

Implements a filesystem and a memory backend to start with.

Fixes #291


rubiojr:
Ready for :eyes:

cfg/config.go
```diff
@@ -1,60 +1,150 @@
...
 // Config contains an entire configuration set for Beehive
 type Config struct {
 	Bees    []bees.BeeConfig
 	Actions []bees.Action
 	Chains  []bees.Chain
+	backend IConfigBackend
+	url     *url.URL
+}
+
+// IConfig is the interface implemented by the configuration backends
```
muesli, line 27: I must admit I chuckled a little when I saw `IConfig`. Let's try and keep using a regular name here, as idiomatic Go suggests. Also, the doc-comment doesn't match the declared type below: `IConfig` vs `IConfigBackend`.

We could call this just `ConfigBackend` or may be something like `ConfigHandler`.

cfg/config_test.go
```diff
@@ -1,71 +1,56 @@
...
 )
 
-func TestLoadConfig(t *testing.T) {
-	_, err := LoadConfig("foobar")
-	if err == nil {
-		t.Error("Loading an invalid config file should return an error")
-	}
-
-	conf, err := LoadConfig(filepath.Join("testdata", "beehive.conf"))
+func TestNew(t *testing.T) {
```
muesli, line 7: We should probably also try initializing with an invalid URL, to verify it returns an error.

cfg/config_test.go
```diff
@@ -1,71 +1,56 @@
...
+		panic(err)
+	}
+	if _, ok := conf.Backend().(*MemBackend); !ok {
+		t.Error("Backend for 'mem:' should be a MemoryBackend")
 	}
+}
 
-	configFile := tmpfile.Name()
-	err = SaveConfig(configFile, testConf)
+func TestSetURL(t *testing.T) {
```
muesli, line 33: Can't harm, but as I see it this is already covered by `TestNew`, as `New` calls `SetURL`?

cfg/filebackend_test.go
```diff
@@ -0,0 +1,84 @@
...
+	u, _ := url.Parse("file://foobar")
+	backend := NewFileBackend()
+
+	_, err := backend.Load(u)
+	if err != nil {
+		t.Error("Loading an non-existing config file should not return an error")
+	}
+
+	// try to load the config from a relative path
+	u, err = url.Parse(filepath.Join("testdata", "beehive.conf"))
```
muesli, line 21: Doesn't this require `testdata/beehive.conf` to be added to the repo?

cfg/config.go
rubiojr: Fixed

cfg/config_test.go
rubiojr: Indeed. Removed.

cfg/filebackend_test.go
rubiojr: Already there: https://github.com/muesli/beehive/tree/master/cfg/testdata

That what you mean?

cfg/config_test.go
rubiojr: This could be trickier, because `url.Parse` accepts pretty much any random string, and to preserve backwards compat I opted for accepting things like `/path/to/my/beehive.conf`, `c:\Program Folder\foobar.conf`, etc.

Do you think we should be stricter here and only accept URLs with a valid scheme, host or path, port, etc?

rubiojr: We should at least clarify that adding some docs and maybe change the parameter name to something else to make the intention clear.

rubiojr: Actually I'm not sure native windows filesystem paths are really accepted by Beehive currently, i.e. `C:\foobar\stuff`. If you run Beehive from master with something like `beehive --config 'c:\foobar\stuff', it'll save the config to `$CWD/C:\foobar\stuff` in a Unix system.

Should we leave better config path handling for a new, dedicated PR where we take into consideration the issues with different OSes?

cfg/config.go
```diff
@@ -20,12 +20,12 @@ type Config struct {
 	Bees    []bees.BeeConfig
 	Actions []bees.Action
 	Chains  []bees.Chain
-	backend IConfigBackend
+	backend ConfigBackend
 	url     *url.URL
 }
 
 // IConfig is the interface implemented by the configuration backends
```
muesli: Should read ConfigBackend here as well.

rubiojr: fixed too.

cfg/config_test.go
rubiojr: Ended up adding some extra docs, better validation and some extra tests.

Let me know if that works.

rubiojr:
Thanks for the review @muesli!

muesli:
Hah, a much bigger thanks for your continuous support and contributions, @rubiojr! Can't express how much I appreciate all your support :rocket:

PR #296, opened 2020-04-16
[PoC] Debian package and systray icon

rubiojr:
A lighter and simpler alternative to something like https://github.com/muesli/beehive/pull/142, Debian/Ubuntu only for now.

@muesli been thinking Electron is an overkill (and mem hungry) for what we need and found this alternative, which is x-platform also apparently.

This is the way it looks on Ubuntu, when and after launching it:

Screenshot from 2020-04-16 20-16-30

Screenshot from 2020-04-16 20-15-31

Screenshot from 2020-04-16 20-19-52

Screenshot from 2020-04-16 20-23-43

Would it be useful to release Debian (and possibly RPM) packages with this tiny wrapper?


rubiojr:

PR #301, opened 2020-04-28
Sunrise/Sunset events hive

rubiojr:
Send sunrise/sunset events for a given city.


bees/sunbee/sunbee.go
```diff
@@ -0,0 +1,155 @@
...
+	bees.Bee
+	city string
+}
+
+var sunsetAlerted = false
+var sunriseAlerted = false
+
+// Run executes the Bee's event loop.
+func (mod *SunBee) Run(eventChan chan bees.Event) {
+	gominatim.SetServer("http://nominatim.openstreetmap.org/")
```
muesli, line 48: Probably should prefer `https` here?

muesli:
Oh yes! I so need this hive!

muesli:
Similar to what the `IRCBee` is doing with its `connected` state, we could emit the current state (after sunrise / after sunset), so it can be accessed by all other chains, even ones not triggered by this bee's events.

See https://github.com/muesli/beehive/blob/master/bees/ircbee/ircbee.go#L261

bees/sunbee/sunbee.go
rubiojr: :+1:

rubiojr:
> Similar to what the `IRCBee` is doing with its `connected` state, we could emit the current state (after sunrise / after sunset), so it can be accessed by all other chains, even ones not triggered by this bee's events.
> 

@muesli TIL, that's pretty awesome. Is abd067ec3d7c2007f90a1fe42978d8fa01c534da what you had in mind?

muesli:
> @muesli TIL, that's pretty awesome. Is [abd067e](https://github.com/muesli/beehive/commit/abd067ec3d7c2007f90a1fe42978d8fa01c534da) what you had in mind?

Yes, that looks about right! Now you should be able to use `.context.beename.sunset` (e.g.) as a value in any action.

bees/sunbee/sunbeefactory.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// Image returns the filename of an image for this Bee.
+func (factory *SunBeeFactory) Image() string {
+	return factory.ID() + ".png"
+}
+
+// Options returns the options available to configure this Bee.
+func (factory *SunBeeFactory) Options() []bees.BeeOptionDescriptor {
+	opts := []bees.BeeOptionDescriptor{
```
muesli, line 64: It might be nice to be able to set an offset for the sunset/sunrise (probably individually).

For example I'd like to use this to turn on the lights at home about half an hour to an hour before the actual sunset.

muesli:
I'm so gonna drop TravisCI and replace it with GitHub Actions. It's still trying to boot up the macOS jobs from several hours ago :yawning_face:

bees/sunbee/sunbeefactory.go
rubiojr: Done!

rubiojr:
Ready for :eyes:

bees/sunbee/sunbee.go
```diff
@@ -36,19 +36,37 @@ import (
...
+		mod.LogFatal("failed parsing longitude from response. err: %v", err)
+	}
 
 	for {
 		select {
 		case <-mod.SigChan:
 			return
-		case <-time.After(time.Duration(1 * time.Minute)):
-			mod.check(mod.city, eventChan)
+		case <-time.After(time.Duration(10 * time.Second)):
```
muesli: Isn't this a bit excessive?

bees/sunbee/sunbeefactory.go
```diff
@@ -63,11 +63,18 @@ func (factory *SunBeeFactory) Image() string {
...
+			Name:        "query",
 			Description: "The name of the city where the sunrise/sunset will happen",
 			Type:        "string",
 			Mandatory:   true,
 		},
+		{
+			Name:        "offset",
+			Description: "Fire the event this number of seconds before sunset/sunrise (default: 2 min)",
+			Type:        "int",
+			Default:     120,
```
muesli, line 75: Maybe we should keep the default offset at 0 and make this non-mandatory? I don't think it needs to be enforced.

bees/sunbee/sunbee.go
```diff
@@ -0,0 +1,168 @@
...
+	"math"
+	"strconv"
+	"time"
+
+	"github.com/kelvins/sunrisesunset"
+	"github.com/muesli/beehive/bees"
+	"github.com/muesli/gominatim"
+)
+
+// SunBee is an example for a Bee skeleton, designed to help you get started
```
muesli, line 35: Comment from HelloBee :smile:

bees/sunbee/sunbee.go
muesli: Ignore that comment, just saw you already fixed that in a later commit and my GitHub diff just wasn't up-to-date.

bees/sunbee/sunbee.go
```diff
@@ -0,0 +1,168 @@
...
+
+// Run executes the Bee's event loop.
+func (mod *SunBee) Run(eventChan chan bees.Event) {
+	gominatim.SetServer("https://nominatim.openstreetmap.org/")
+	qry := gominatim.SearchQuery{
+		Q: mod.query,
+	}
+	resp, err := qry.Get()
+	if err != nil {
+		mod.LogFatal("Error geocoding %s. err: %v", mod.query, err)
```
muesli, line 52: Careful, I think `LogFatal` doesn't support formatting directives.

bees/sunbee/sunbeefactory.go
rubiojr: Done. Bear in mind that the check interval is 60 secs so we'll always have a +/- 60 secs offset/error.

bees/sunbee/sunbee.go
rubiojr: Thanks, fixed!

rubiojr:
I think I've addressed all your feedback @muesli

PR #302, opened 2020-04-30
Adds TimeNow template helper

rubiojr:
So you can use something like `{{TimeNow.Format "15:04:05"}}`, to add
timestamps to your actions.


rubiojr:
@muesli is there a way to do this now that I missed?

rubiojr:
@muesli cool, been using it, seems to work as intended. Do you want me to update the Wiki?

muesli:
Yes please! :heart:

PR #305, opened 2020-05-01
Add flag to decrypt the configuration

rubiojr:
Needs https://github.com/muesli/beehive/pull/295 to be merged first.

PR #307, opened 2020-05-01
Add flag to decrypt the configuration

rubiojr:
Attempt two at https://github.com/muesli/beehive/pull/305 that was merged to the wrong branch accidentally.


rubiojr:
Fixed the Debugf you also pointed out @muesli.

PR #309, opened 2020-05-07
Embedding API

rubiojr:
Highlights:
* Exposes a new `reactor` package (not particularly attached to this name) .
* Doesn't change existing API.

This makes it easier to:
* Embed Beehive's in other binaries or libraries.
* Create custom Beehive binaries with additional hive packages hosted elsewhere.
* Create specialized binaries which benefit from Beehive's vast catalog of bees, making it possible to create really powerful interactions with a few lines of code.

Clients embedding Beehive's don't have a built-in HTTP API server started by default, which is probably desirable and safer if you're programming your own chains, as opposed to using the GUI to create them.

[I've included an example](https://github.com/muesli/beehive/blob/a07430c277542b970438b48b6c921287b3534deb/examples/embedding/main.go) that demonstrates embedding. Also implemented the Hello bee action the example is using.

In future iterations, we could expose an interface to start/stop the API server also, maybe add a few functions that wrap the event, action and chain creation while adding some safety.


rubiojr:
A slightly more complex example that demonstrates using an out-of-tree bee in an external repo: https://github.com/rubiojr/commit-monitor

rubiojr:

PR #310, opened 2020-05-09
MS Windows fixes

rubiojr:
* Handle Windows path arguments like `c:/path/to/beehive.conf` when using --config.
* Handle crypto URLs correctly
* Test fixes, making sure we always use forward slashes in paths under Windows.
* `c:\foo\bar` windows paths aren't accepted (use forward slash instead)

For some extra context, Go URL parsing in Windows can be a bit surprising.

Parsing `file:///c:/beehive.conf`, the URI format documented [here][1],
will set the URL path to `/c:/beehive.conf`.
This behaviour is captured in https://github.com/golang/go/issues/6027.

If we use a URL like file://c:/beehive.conf, the drive name will be parsed
as the host and the path will miss the drive name.

```go
package main

import (
        "fmt"
        "net/url"
)

func main() {
        u, _ := url.Parse(`file://c:/beehive.conf`)
        fmt.Println(u.Host)
        fmt.Println(u.Path)
}
```

This prints:

```
c:
/beehive.conf
```

We workaround it patching the URL path and host so URLs like
`crypt://sercret@c:/beehive.conf` behave as expected.

[1]: https://docs.microsoft.com/en-us/archive/blogs/ie/file-uris-in-windows


rubiojr:
I think I can come up with a slightly cleaner fix so I don't have to patch every single backend method.

rubiojr:
@muesli if you don't entirely dislike this patch, we could merge as is to fix the windows situation and I'll work on a new PR, to clean things up a bit.

beehive.go
```diff
@@ -106,7 +105,7 @@ func main() {
 		log.Fatalf("Error creating the configuration %s", err)
 	}
 
-	if config.URL().String() != cfg.DefaultPath() { // the user specified a custom config path or URI
```
rubiojr, line 109: String() does some encoding when using Windows paths which is not what we want here.

rubiojr:
This is looking good now.

Ended up wrapping `net/url` `URL` with a drop-in replacement that helps to encapsulate our platform specific logic to patch URLs in Windows.
The wrapper is heavily exercised buy our current tests so I didn't consider necessary to add dedicated unit tests for it.

Also took the chance to add some test helpers and clean things up a bit.

rubiojr:
> This is looking good now.

Not yet, just found one more windows niggle when executing the binary.

rubiojr:
Fixed in https://github.com/muesli/beehive/pull/310/commits/602668adc400c0a5ef2465d1fe1080ef1099e060

muesli:
@rubiojr Just a heads up, @penguwin and I have been pondering how to handle the config URL situation properly for a while and have come up with a few fixes. I think we'll probably have to combine our approaches.

rubiojr:
sounds good!

penguwin (commented):
Hey @rubiojr o/

Sorry for responding this late - I've somehow completely missed your PR on this. 

I implemented a configuration system in [knoxite](https://github.com/knoxite/knoxite) based on your awesome work in beehive. I'm also about to contribute back some tweaks and enhancements I made in knoxite to beehives config system. As @muesli already addressed I also think it's best to somehow combine both systems - or at least don't let them diverge that much - so we can push improvements and fixes to both programs. 

In knoxite the same issues regarding windows paths occurred, however we've used another approach to resolve them. We wrapped a method called [PathToUrl](https://github.com/knoxite/knoxite/blob/be35bb0bcaa111f06dac7091822c4e8b8853b9df/cmd/knoxite/utils/utils.go#L161) around `url.Parse()`, called by the configs `setURL()` method. This method *should* be able to parse all kind of different formatted config file paths (unix/windows) without converting backslashes with a `fixWindowsPath` method.

rubiojr:
Awesome @penguwin, let me have a look at the implementation there and I'll get back to you ASAP.

penguwin:
cool, if there's anything I can help/assist you with let me know

rubiojr:
@penguwin: rebased the branch but not without some :sweat_smile:, since I forgot what I did here in May :smile: .

It'll need some extra polish before asking for a reviewing it again, so I'm taking care of that first.

If y'all haven't started it, I'll also be looking into extracting the configuration system into it's own repo, see if we can make that work for both beehive and knoxite, to avoid duplicated work and the tedious maintenance task of cherry picking fixes for it from one project to the other.

rubiojr:

PR #312, opened 2020-05-18
openweathermap: Handle multiple locations in context

rubiojr:
Extends https://github.com/muesli/beehive/pull/311 by adding support
to store multiple locations in the Bee context, at the expense of making
it a bit more complex to retrieve the info from a template.

Something like:

```
"Weather in {{(index .context.weather.current 3128760).Name}}: ..."
```

Where `3128760` is the OpenWeatherMap ID corresponding to one of the
locations stored in the context while retrieving the weather.

### DIFF
--- a/bees/openweathermapbee/event.go
+++ b/bees/openweathermapbee/event.go
@@ -27,10 +27,19 @@ import (
 	owm "github.com/briandowns/openweathermap"
 )
 
+type weatherContext = map[int]*owm.CurrentWeatherData
+
 // TriggerCurrentWeatherEvent triggers all current weather events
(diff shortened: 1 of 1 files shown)


rubiojr:
Kinda torn about this one. Makes it possible to retrieve the weather of multiple locations from the context but also makes it uglier to use it from a template.

The alternative would be to create multiple weather bees for each location, which is currently supported by the previously merged PR, https://github.com/muesli/beehive/pull/311

muesli:
I agree: our typical approach here would be to initialize multiple instances of the `OpenWeatherMapBee`, unless there are other advantages to sharing a single instance.

PR #323, opened 2020-06-16
Adds multi-arch docker-build to release script

### LINKED ISSUE #314: Publish arm docker container as well as amd64
Should hopefully be simple enough; it builds just fine using the existing Dockerfile (to https://hub.docker.com/repository/docker/cerebrate/beehive , for example), but it would be good to have an official image that keeps pace with releases.

rhnvrm:
Closes #314


dereulenspiegel:
Any chance that this gets merged soon? Would really love to run beehive on a Raspberry Pi :)

muesli:
Afraid that's only part of a solution. Ideally we integrate Docker builds with goreleaser, as this release script is a bit obsolete. Sadly there are still issues with goreleaser and cross-platform builds.

We'll also have to disable the automated Docker hub builds, otherwise these approaches will constantly race each other, at least for the `latest` tag.

I will do a manual build & push for the latest stable release, so you can at least start using it on a Pi.

muesli:
Actually this is even more complicated than anticipated, because buildx is still an experimental feature and not enabled by default.

muesli:
This might help us: https://goreleaser.com/customization/docker_manifest/

rhnvrm:
Hey @muesli 

I implemented the github action (ref: https://github.com/muesli/beehive/issues/314#issuecomment-723153160).

> This might help us: https://goreleaser.com/customization/docker_manifest/

It uses the same experimental setup, so even if we use goreleaser, you will have to set up `buildx` locally. To avoid this local setup, we can continue using goreleaser to create the tag and push it to the repo; on push of the tag, the github action will build and push the docker image. (ref this PR: https://github.com/muesli/beehive/pull/323/commits/c2aca9d244ad1ff4ec39daa1ffed56e265cedbcb)

Issue #293, opened 2020-04-13
Upgrade documentation for the cron and timer hives


Zykino:
I just saw there are Cron and Time hives and they look too be identical.
The descriptions should better state that Cron is recurring and Time one time only ? The Time hive may need a "once" or "one time only" in the description to make the difference really clear.

And if possible they should be near (Time once, Time recurring (cron)) so user will see both at once and choose whichever they need.

rubiojr:
:+1: The intention of both hives hasn't been clear to me either. I've been thinking we had two different implementations that do the same thing for a while.

rubiojr:
@muesli, would it help to rename that hive to "Once", "On This Date" or something like that?

Zykino:
I heard you are reworking the admin interface (potentially adding categories) but I think that it is quite important that since this two hives are so close (utility, logo, input, descriptions) that they need to share part of the name and be differentiated on an other part. That way they will be next to each other in the list and as a user I will ask myself which one I need right now. Except if you intent to randomize the hives order…

Something like:
Time (Once)
Time (Recurrent)

muesli:
I agree, we should clean that up and combine those two Hives into one, which then offers multiple event types.

We should probably also drop the underlying cron library and switch to a maintained and better tested one: https://github.com/robfig/cron

PR #327, opened 2020-07-24
Use robfig/cron to parse cron lines

rubiojr:
Handles parsing better, supporting the cron spec.

Partially addresses https://github.com/muesli/beehive/issues/293


rubiojr:
This is ready for :eyes: now. I ran `go mod tidy` so the diff is a bit larger than necessary.

rubiojr:
The format supported is documented in https://godoc.org/github.com/robfig/cron.

```
Field name   | Mandatory? | Allowed values  | Allowed special characters
----------   | ---------- | --------------  | --------------------------
Minutes      | Yes        | 0-59            | * / , -
Hours        | Yes        | 0-23            | * / , -
Day of month | Yes        | 1-31            | * / , - ?
Month        | Yes        | 1-12 or JAN-DEC | * / , -
Day of week  | Yes        | 0-6 or SUN-SAT  | * / , - ?
```

Tested with an old config I had around and seems to be backwards compatible. It seems to fix some issues the current parse has also, like no support for `*/X` periods.

muesli:
Fantastic!

PR #330, opened 2020-08-26
Add bash and python2 for CMD hive usage

rocket357:
The default beehive docker image shell (busybox) leaves a lot to be desired when it comes to executing commands via the CMD hive.  This PR adds bash and python2 to the installed packages to allow them to be used in CMD.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,7 @@ RUN make embed
 FROM alpine
 
 RUN apk update && \
-    apk add --no-cache ca-certificates tzdata && \
+    apk add --no-cache ca-certificates tzdata bash python2 && \
     update-ca-certificates
 
 COPY --from=builder /go/beehive/beehive /go/bin/beehive


rocket357:
There is a "bit" of size difference (as expected):

`root@beehive:~# docker image list | grep beehive
beehive                        latest              b70912e0ab24        6 days ago          71.7MB
fribbledom/beehive             latest              4c5db390fbc9        2 weeks ago         33.5MB
`

muesli:
I'm a bit torn on this one. Bash probably makes sense, but where do we draw the line with regards to python?

rocket357:
Yeah, python is quite arbitrary (both version and python itself), and is
included here because my specific requirement is for python2, and not, say,
ruby. Would it make more sense to write up documentation on customizing the
docker image, or perhaps providing both a minimal and a "kitchensink"
docker image?

muesli:
> Yeah, python is quite arbitrary (both version and python itself), and is included here because my specific requirement is for python2, and not, say, ruby. Would it make more sense to write up documentation on customizing the docker image, or perhaps providing both a minimal and a "kitchensink" docker image?

I like **both** these ideas a lot!

rocket357:
Is there a way to gather metrics (automated preferably, but a poll of beehive users would work, too, if not) to see what everyone is using with command bees?  If polling, do you have preferences on how/who/when/etc...?

rocket357:
Side note: Seeing build/test (1.15.x ubuntu-latest) fail with:

Error: bees/notificationbee/notificationbee_unix.go:43:19: conversion from uint32 to string yields a string of one rune, not a string of digits (did you mean fmt.Sprint(x)?)

rocket357:
First run test image (currently testing on my kubernetes cluster):

https://hub.docker.com/r/rocket357/beehive-kitchensink

Based on:

https://github.com/rocket357/beehive-kitchensink/blob/main/Dockerfile

Perhaps we can configure a separate branch to automate building on Dockerhub?  That may introduce some sync issues on the github side.  Any ideas on how best to tackle getting a separate Dockerfile into fribbledom/beehive official?  We could use a sub folder in github and build rules on Dockerhub to automate, I suppose.

rocket357:
Since dockerhub has removed build automation for free accounts, I've set up a build server to automate building the beehive-kitchensink images ([available here](https://hub.docker.com/r/rocket357/beehive-kitchensink)).  I've removed python2 since it is well into EOL and I'm building multi-arch images (currently arm32v7, arm64v8, and amd64).

PR #333, opened 2020-10-02
Test template filters

orsinium:
1. Make template filters kind of type-safe. leave type reflection to the template engine and accept already the right type. Also, it helps to correctly match the number of arguments, making error messages in that case user-friendly.
2. Add unit tests for happy-path for custom filters.
3. Do not suppress errors from filters, making them propagate up into the template engine. The engine gracefully handles it and returns as the second argument from `Template.Execute` method indicating template execution failure.

Thank you for the amazing project! I'm migrating from IFTTT :)


orsinium:
CI fails not because of my changes, I'm sure. Is it flaky? :thinking:

rocket357:
Yes. I put in a PR for docker and CI failed for windows! The docker file
isnt (to my knowledge) used or touched in the CI builds. So I would assume
CI is flakey.

muesli:
Just a heads up: @penguwin is already working on fixing the Windows builds.

muesli:
Great contribution, thank you!

PR #334, opened 2020-10-02
Support parse_mode for telegram

orsinium:
Telegram supports [formatting options](https://core.telegram.org/bots/api#formatting-options) which can be specified via `parse_mode` field. It is a must-have thing to have nice messages and `telegram-bot-api` supports it.

The PR makes it possible to specify `parse_mode` when configuring Telegram Action.


muesli:
Great contribution, thank you!

PR #336, opened 2020-10-05
Add a few helpful template filters

orsinium:
## ContainsAny

I patched `ContainsAny` filter to be able to check if the given string contains any of the given substrings.

For example, this one:

```clojure
{{test or (Contains .title "dephell") (Contains .title "flakehell") (Contains .title "orsinium.dev")}}
```

Can now be transformed into much simpler and shorter expression:

```clojure
{{test ContainsAny .title "dephell" "flakehell" "orsinium.dev"}}
```
If only 2 strings are passed into the filter, it behaves as it used to, checking every symbol from the second string to be represented in the first one. While I cannot imagine a use case for it in beehive, let's keep the backward compatibility. It should be ok because `ContainsAny` looking only for one substring would be the same as just `Contains`.

## ContainsAll

If we have `ContainsAny`, why not to have `ContainsAll`? I've added a new filter to check if a string contains all substrings.

## Array

There is no syntax in the template engine to create an array but some functions accept arrays. It can be emulated with `Split` but it's not nice looking. So, let's make one.

```clojure
{{Array "flakehell" "dephell"}}
```

## More tests

I've added more tests for string functions. Since there is no way to edit the wiki from the master branch (where is it, BTW?), let's at least have the code "self-documented".

### DIFF
--- a/templatehelper/templatehelper.go
+++ b/templatehelper/templatehelper.go
@@ -75,15 +75,28 @@ var (
 			return items[len(items)-1], nil
 		},
 		"ContainsAny": func(target, subs string, other ...string) bool {
+			// contains any symbol from the given string
 			if len(other) == 0 {
 				return strings.ContainsAny(target, subs)
 			}
+			// contains any of the given strings
 			for _, another := range other {
 				if strings.Contains(target, another) {
 					return true
 				}
 			}
 			return strings.Contains(target, subs)
+		},
+		"ContainsAll": func(target string, other ...string) (bool, error) {
+			if len(other) == 0 {
+				return false, errors.New("to find substring in a string, use Contains")
+			}
+			for _, another := range other {
+				if !strings.Contains(target, another) {
+					return false, nil
+				}
+			}
+			return true, nil
 		},
 		// strings functions
 		"Compare":      strings.Compare, // 1.5+ only
--- a/templatehelper/templatehelper.go
+++ b/templatehelper/templatehelper.go
@@ -34,6 +34,9 @@ import (
 // FuncMap contains a few convenient template helpers
 var (
 	FuncMap = template.FuncMap{
+		"Array": func(values ...interface{}) []interface{} {
+			return values
+		},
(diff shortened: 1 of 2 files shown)


muesli:
> If only 2 strings are passed into the filter, it behaves as it used to, checking every symbol from the second string to be represented in the first one. While I cannot imagine a use case for it in beehive, let's keep the backward compatibility. It should be ok because `ContainsAny` looking only for one substring would be the same as just `Contains`.

I really don't like this kind of ambiguity. Why not introduce a separate function for that behavior?

> ## Array

Nice!

> ## More tests

Awesome!
 
> I've added more tests for string functions. Since there is no way to edit the wiki from the master branch (where is it, BTW?), let's at least have the code "self-documented".

GitHub wikis are in a separate git repository: https://github.com/muesli/beehive.wiki.git

orsinium:
> I really don't like this kind of ambiguity. 

Me too. However, the new behavior (check if any of substrings is in the string) is what I would expect from `ContainsAny`. I'd remove the old behavior because `ContainsAny "something" "abc"` is the same as `ContainsAny "something" "a" "b" "c"` from now on but it would be a breaking change.

> Why not introduce a separate function for that behavior?

How would you name it?

muesli:
I can't come up with anything better than `ContainsAnyString` :disappointed: Still open for suggestions :smiley:

orsinium:
A few suggestions:

1. Accept an array of strings as the second argument for the new behavior instead of variadic args.
1. Screw the backward compatibility, return an error if only 2 arguments were passed like it is done in `ContainsAll`. The old implementation of `ContainsAny` rename to `ContainsAnyRune`.
1. Don't touch `ContainsAny`, extend `Contains` instead.
1. Don't worry about mixing behaviors (the implementation in this PR).

orsinium:
Does it make sense for me to keep the PR open any longer or should I just close it? It's ok to not want to merge some random contributions. I'm long enough in open source to understand :)

PR #338, opened 2020-10-05
Fix go-lint violations in `cfg` package

orsinium:
I've run `golangci-lint` on `cfg` package and it found some violations. Mostly, it comes from `go-lint` and about ignored variables and unchecked errors.

The PR fixes all the violations.

### DIFF
--- a/cfg/config_test.go
+++ b/cfg/config_test.go
@@ -26,6 +26,9 @@ func TestNew(t *testing.T) {
 	cwd, _ := os.Getwd()
 	p := filepath.Join(cwd, "testdata/beehive-crypto.conf")
 	conf, err = New(p)
+	if err != nil {
+		panic(err)
+	}
 	if _, ok := conf.Backend().(*AESBackend); !ok {
 		t.Errorf("Backend for '%s' should be an AESBackend", p)
 	}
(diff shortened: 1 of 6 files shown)


cfg/config_test.go
```diff
@@ -9,15 +9,17 @@ import (
 func TestNew(t *testing.T) {
 	conf, err := New("/foobar")
 	if err != nil {
-		panic(err)
+		t.Error("cannot create config from path")
```
muesli, line 12: Couldn't we just use `t.Fatal` here?

cfg/config_test.go
```diff
@@ -26,24 +28,28 @@ func TestNew(t *testing.T) {
 	cwd, _ := os.Getwd()
 	p := filepath.Join(cwd, "testdata/beehive-crypto.conf")
 	conf, err = New(p)
+	if err != nil {
+		panic(err)
```
muesli, line 32: I think you just forgot that one.

cfg/config_test.go
orsinium: Thank you! I'm a bit lost without assertify :upside_down_face: Fixed

cfg/config_test.go
orsinium: Great catch! Fixed.

orsinium:
Review comments are applied :+1:

muesli:
Thank you @orsinium, loving all the PRs!

PR #340, opened 2020-10-07
Improve filters interface

orsinium:
`execFilter` has the following signature: `execFilter(filter string, opts map[string]interface{})`.. Values of `opts` and `filter` are passed as-is into  `FilterInterface.Passes`.

The PR adjusts `FilterInterface` to reflect actual types if `opts` and `filter` instead of a general `interface{}`.

### DIFF
--- a/filters/template/templatefilter.go
+++ b/filters/template/templatefilter.go
@@ -45,28 +45,23 @@ func (filter *TemplateFilter) Description() string {
 }
 
 // Passes returns true when the Filter matched the data.
-func (filter *TemplateFilter) Passes(data interface{}, value interface{}) bool {
-	switch v := value.(type) {
-	case string:
-		var res bytes.Buffer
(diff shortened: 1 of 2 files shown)


filters/template/templatefilter.go
```diff
@@ -45,28 +45,23 @@ func (filter *TemplateFilter) Description() string {
...
-		}
-
-		tmpl, err := template.New("_" + v).Funcs(templatehelper.FuncMap).Parse(v)
-		if err == nil {
-			err = tmpl.Execute(&res, data)
-		}
-		if err != nil {
-			panic(err)
-		}
+	if strings.Contains(v, "{{test") {
```
orsinium, line 51: Also, here I changed the `strings.Index` to `strings.Contains` because it is the same but is faster. Thanks to `go-simple`.

orsinium: (everything else here is just changed identation)

muesli:
Thank you!

PR #341, opened 2020-10-07
Starlark

orsinium:
Add [Starlark](https://github.com/google/starlark-go) template filter syntax support. See the included documentation for more information

## Motivation

1. More powerful: everything that we can do with "template" filters can be done with "starlark" and much more. Actually, Starlark is turing-complete.
1. More readable.
1. Familiar: Python is one of the most popular languages.

## Requirements

Blocked by #340 (contains that PR in the diff)

## TODO

The `input` field for filters in [beehive-admin](https://github.com/muesli/beehive-admin) should be changed to `textarea`. For now, I'm using YAML configuration file from #339 to write starlark filters.

### DIFF
--- a/bees/filters.go
+++ b/bees/filters.go
@@ -43,14 +43,18 @@ func execFilter(source string, opts map[string]interface{}) bool {
 		name = "starlark"
 	}
 
-	f := *filters.GetFilter(name)
+	filter := filters.GetFilter(name)
+	if filter == nil {
+		log.Error("cannot find filter", name)
+		return false
+	}
 	log.Println("\tExecuting filter:", source)
 
 	defer func() {
 		if e := recover(); e != nil {
-			log.Println("Fatal filter event:", e)
+			log.Error("Fatal filter event:", e)
 		}
 	}()
 
-	return f.Passes(opts, source)
+	return (*filter).Passes(opts, source)
 }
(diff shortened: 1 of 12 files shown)


bees/filters.go
```diff
@@ -35,15 +37,24 @@ type Filter struct {
 }
 
 // execFilter executes a filter. Returns whether the filter passed or not.
-func execFilter(filter string, opts map[string]interface{}) bool {
-	f := *filters.GetFilter("template")
-	log.Println("\tExecuting filter:", filter)
+func execFilter(source string, opts map[string]interface{}) bool {
+	name := "template"
+	if strings.Contains(source, "def main(") {
```
muesli, line 42: I'm not too happy with this "string magic". Maybe we could introduce proper filter types? I was also pondering having simple bash-scripts as filters, which either return 0 or 1 as a result.

orsinium: I'm also not sure if it is a good solution but it requires a selector on the UI side and I'm not ready to dive into Ember. If you're ready to do such adjustments, please, do.

If to talk about a better backend-only solution, I can propose shebangs. In that case, this line will be `if strings.HasPrefix(source, "# starlark") `, for example.

Bash scripts is a cool and fun idea. I'm sure some will be happy to build grep/sed/awk/perl pipelines.

Issue #264, opened 2019-10-17
Facebook bee setup problem


Invertisment:
I wanted to fiddle with facebook bee and before I could even try it out I got an error.
It's related to the deprecated post API.

I am in no hurry to fix anything and this doesn't impact me, but it would be nice for it to work.

https://www.nextscripts.com/news/2018/04/facebook-error-invalid-scopes-publish_actions/

CalmBit:
Could you post the error here? I'd like to debug this and see what's going on.

Invertisment:
I've tried again and this time I noticed that I didn't set my fb app to "live" (I haven't created any fb apps too).
With this being out of the way it could work but now I get an error about the domain of the app (in lithuanian):

Nepavyksta atverti nuorodos: Nuorodos domenas nėra įtrauktas į programėlės domenų sąrašą. Programėlės nustatymuose, lauke Programėlės Domenai, pridėkite visus programėlės domenus ir subdomenus, ir jūs sėkmingai galėsite atidaryti šią nuorodą.

As I understand this basically means that I have to add the domain of the app to the `Domains`, which I already tried for a couple of times. And probably as I've tried to make it work with `localhost` and `127.0.0.1` they may not work at all.

As I've doing this only for exploration I would like to have any tutorials that may be useful. Probably it would be a nice addition. 
Thanks.

CalmBit:
[This](https://wp-native-articles.com/blog/news/how-to-fix-facebook-apps-error-cant-load-url-domain-url-isnt-included-apps-domains/) seems to cover this error pretty well. I'll have to research it more, but the HTTPS requirement might be the first place to start.

pbek:
> Invalid Scopes: publish_actions. This message is only shown to developers. Users of your app will ignore these permissions if present. Please read the documentation for valid permissions at: https://developers.facebook.com/docs/facebook-login/permissions

I did get the same error message when trying to retrieve an access token.
The facebook hive tries to get `publish_actions` permission. This permission doesn't seem to exist any more.
I cannot find them on https://developers.facebook.com/docs/facebook-login/permissions or in the permission list on https://developers.facebook.com/tools/explorer/.

pbek:
I valid permission could be [pages_manage_posts](https://developers.facebook.com/docs/permissions/reference/pages_manage_posts).

pbek:
I replaced the permissions in Beehive which allowed me to get an access token, but when I tried to post a message to my Facebook page I got an error:

```
[...]

PR #345, opened 2020-10-24
Attempt to fix Facebook auth scopes (#264)

pbek:

PR #346, opened 2020-10-25
Attempt to fix Facebook auth scopes (#264)

pbek:


pbek:
Follow up from #345.

> Have you tried authing with Facebook with the new app scope?

I now have setup my own build and docker deployment process so I can test on my server.
After the painful configuration at https://developers.facebook.com (and successifuly receiving an auth token) I'm one step further.

```
ERRO[0993] [Facebook 2]: Error: [message:(#200) If posting to a group, requires app being installed in the group, and \
          either publish_to_groups permission with user token, or both pages_read_engagement \
          and pages_manage_posts permission with page token; If posting to a page, \
          requires both pages_read_engagement and pages_manage_posts as an admin with \
          sufficient administrative permission] [type:OAuthException] [code:200] [subcode:0] 
```

So yes, the scopes are working.

pbek:
@muesli, is the Facebook Hive capable of posting to a Facebook page? Because that's what I'm trying to achieve. 😀

pbek:
It seems like pages are a different thing, you need a page access token, https://developers.facebook.com/docs/pages/publishing.

muesli:
I must admit I've kinda stopped using Facebook personally, but have so far only used it to post on my personal page - which Facebook now doesn't permit anymore.

Posting to pages via the API should still be doable, though. We may or may not have to adapt to their newer API versions.

pbek:
Facebook posting was always kind of flaky (back a few years ago when I wrote my own publishing integration) since access tokens seem to expire from time to time...

muesli (approved):

muesli:
Thank you, @pbek!

Issue #335, opened 2020-10-03
100% CPU usage if using Twitter bee


Phoenix616:
It seems to be using 100% of one core/thread once I enable a Twitter bee even without using any kind of chain which would react on twitter actions (I only use it to send tweets)

muesli:
Thanks for the heads up. That may be related to Twitter shutting down several API endpoints. Investigating.

pbek:
@muesli, I was able to post to Twitter, but two cores of my server are at 100%! Were you already able to take a look at the Twitter Hive? Reacting to Twitter likes didn't seem to work for me, `Last Event` always says `never` for me.

pbek:
@muesli, the only thing that runs infinitely in the Twitter Bee seems to be:

https://github.com/muesli/beehive/blob/f6ac3ac98e38727fdf0255f85ee26ab4298c3c7a/bees/twitterbee/twitterbee.go#L290-L295

pbek:
Although `item` seems to be `nil` a lot (all the time?).

pbek:
I was able to fix the 100% CPU problem, but reacting to a like still doesn't seem to work for me. At least posting tweets is still working...

https://github.com/muesli/beehive/pull/347

pbek:
Last commit on https://github.com/ChimeraCoder/anaconda (library that handles the Twitter API) was 2 years ago. 😬

pbek:
The user stream <https://userstream.twitter.com/1.1/user.json> is [retired since more than 2 years](https://developer.twitter.com/en/docs/twitter-api/v1/accounts-and-users/subscribe-account-activity/migration/us-ss-migration-guide).

The replacements seem to be the "Premium Account Activity API" and "Enterprise Account Activity API". 😁  At least posting tweets still seems to work.

muesli:
Yeah, I guess most people stopped using the Twitter API at that point, myself included. I'm actually one of the anaconda maintainers, but development has pretty much ceased when Twitter shut down most of their API for public use.

pbek:
Posting still works (and I plan to use it), only the user stream is shut down.

muesli:
I've just removed the user stream from the Twitter hive, as it's not supported by Twitter anymore. This in turn means the TwitterBee now only supports posting, following & unfollowing other users. There are no more events we can currently support. What a shame.

This also fixes this issue.

pbek:
[...]

PR #347, opened 2020-10-25
Fix Twitter 100% CPU (#335)

pbek:
This fixes the 100% CPU problem of https://github.com/muesli/beehive/issues/335.

- Posting a tweet still works
- Reacting to a like still doesn't work

### DIFF
--- a/bees/twitterbee/twitterbee.go
+++ b/bees/twitterbee/twitterbee.go
@@ -292,6 +292,7 @@ func (mod *TwitterBee) handleStream() {
 			return
 		case item := <-s.C:
 			mod.handleStreamEvent(item)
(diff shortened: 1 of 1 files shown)


muesli:
I'm not sure that's the right fix: the sleep certainly prevents it from using all of your CPU, but it looping constantly means there's something else going wrong. It's probably because the user-stream API has pretty much been shutdown by Twitter. I guess we can remove it from Beehive altogether now.

pbek:
Posting still works! see #335

muesli:
Oh sorry, I didn't mean to remove the Twitter hive entirely. I only meant to remove the user stream from the code.

muesli:
Replaced by #348.

PR #348, opened 2020-10-26
Remove user stream and events from TwitterBee

muesli:
Twitter has shut down that API, nothing we can do here.


pbek:
Great, thank you!

PR #358, opened 2020-11-11
Introduce skip_first_allow_newest setting for RSS bee

pbek:
@muesli, this would allow the user to tell the bee to skip all current rss items, but allow the newest item to get through once. This way you are able to try to send posts that were triggered by RSS again if there was a problem with the destination bee.


pbek:
Thank you!

pbek:
I updated https://github.com/muesli/beehive/wiki/RSS.

muesli:
No, thank you @pbek! :smiley:

Issue #344, opened 2020-10-24
Email hive doesn't work with unauthenticated smtp servers


pbek:
I'm using a docker container only beehive has access to as smtp server to send myself emails, but the admin interface doesn't allow you to skip user credentials.

(Thank you for this great project, I want to automate some things with it for https://github.com/pbek/QOwnNotes.)

pbek:
I found out you can get around it by entering some text and removing it again later. That way it works.

muesli:
Ok, we should probably mark these options as non-mandatory.

corrantho:
I think this issue can now be closed.

pbek:
Nice, thank you!

PR #365, opened 2020-12-23
emailbee: added support for unauthenticated SMTP servers

corrantho:
To solve issue https://github.com/muesli/beehive/issues/344.

Username and password are now optional.
If empty, the mail is sent with NoAuth mode.


muesli:
Thank you!

Issue #10, opened 2014-06-25
Jira Bee


muesli:
A bee that triggers events on Jira changes.

teo:
JIRA exposes an RSS feed for the Activity stream. Could it be used for this?

muesli:
Should be easy enough, yeah. A proper plugin would still be nicer :-)

dschmidt:
teo and I discussed it already, we'll work on this together next week when I'm in Augsburg. He already found some go client library which we are going to use probably :+1:

Horrendus:
@dschmidt @teo Do you still remember which go client library for JIRA you found or do you have any recommendations?

dschmidt:
No, obviously we didnt get to it :D

internetimagery:
I would love to see this happen as well. Perhaps not just for triggering on jira events, but also triggering jira from other events. Transitioning issues when other services emit events is a powerful workflow automation, and setting it up with the simplicity offered here would be awesome.

+1 for JIRA at somepoint in the future. :)

corrantho:
I think we can close this issue.

PR #366, opened 2020-12-29
Added a new bee for JIRA integration

corrantho:
As requested in https://github.com/muesli/beehive/issues/10, here is a first version of a new bee for JIRA integration. For the moment it has only one action **create_issue** but **transition_issue** is coming soon.

Since I'm new to Golang and on this project, I would like to have already a code review on this small increment to be sure I'm in line with your development guidelines.

Also, regarding the triggering events, I was thinking about exposing an HTTP endpoint in order to receive JIRA events in real-time via a Webhook. Do you have any constraint with this? I've not seen any other bee using this mechanism (except the WebBee).

Another question regarding the logo. Where do you get them? Should it be an opensource logo?

Thanks in advance.


bees/jirabee/jirabeefactory.go
```diff
@@ -0,0 +1,172 @@
...
+			Name:        "password",
+			Description: "Password or API Token (for the cloud version) used to access the JIRA API",
+			Type:        "string",
+			Mandatory:   true,
+		},
+	}
+	return opts
+}
+
+// Events describes the available events provided by this Bee.
```
muesli, line 93: Just a heads up: you can omit empty `Events` methods entirely, if you don't provide any events.

muesli:
Nice, good start for a new bee. One thing is still missing tho: an icon.

Eventually, it would be great to also provide a couple of events, of course 😊

corrantho:
@muesli thanks for the code review.

I'm going to add some events that's why I kept the `Events` method 😄 
Regarding the logo/icon, should it be opensource or can I just copy the one from the JIRA website?

muesli:
I consider logo usage a grey area. So far we haven't had issues with that, so I'd say let's use the Jira logo. People should be able to recognize it - which is the entire point of having an "icon" :smiley:

corrantho:
I will add one more event to handle issue status updates and I think I will be done with this bee.

corrantho:
@muesli I think it's ready for the final review 😃 
Could you also suggest where to put the documentation?

bees/jirabee/events.go
```diff
@@ -0,0 +1,143 @@
...
+	return jiraEvent, nil
+}
+
+func (mod *JiraBee) handleIssueCreatedEvent(data *JiraEvent) error {
+
+	key := ""
+	summary := ""
+	description := ""
+
+	if data.Issue != nil {
```
muesli, line 66: Wouldn't this indicate an error? Maybe instead of sending out an event with empty data we could just return an error here.

corrantho: I totally agree. Thanks for pointing this out.
I pushed the following update: 
- handleIssueCreatedEvent raises an error when Issue is nil
- handleJiraEvent handles errors

muesli:
LGTM, thank you for your contribution!

Issue #356, opened 2020-11-10
Mastodon Triggers Actions inconsistencies


Sakrecoer:
Hi!

Thanks for an awesome tool!

I have been experimenting with the mastodon bees and actions and found an inconsistency and an unexpected behavior with  `{{.username}}`. 
For reblog action, the definition states: _"Mastodon handle of the user that re-blogged your toot"_
For follow action, the definition states: _"Mastodon handle of the user which triggered the follow event"_

**1. Inconsistency**

Expected result: 
As stated by the definitions of the action

Result:
When the action is _follow_, `{{.username}}` prints the display name of the user following our account.
When the action is _reblog_, `{{.username}}`  prints the display name of our account.

**2.  Unexpected behavior (potentially wish-list)**

Expected result
`{{.username}}` prints the users handler: `@user@example.com`

Result
`{{.username}}` prints the user's display name:  _🦄 User's cool display name 🎉_

While it is interesting to access the display name, it would be even nicer to get the handler `@user@example.com` because of how actionable it is compared to the display-name. This is what i expected the role of `{{.user_id}}` to be, but the latter seems to print the digits associated with the user (ex: 765432).

muesli:
Thanks @Sakrecoer, I'll look into this!

corrantho:
In `beehive/mastodonbee/events.go` we have the following code (line 177):

``` 
				{
					Name:  "username",
					Value: notif.Account.DisplayName,
					Type:  "string",
				},
``` 

The event field `username` is indeed mapped with `notif.Account.DisplayName`.
However, here is an extract of the information returned by the Mastodon GET ACCOUNT API (https://docs.joinmastodon.org/methods/accounts/#retrieve-information):

```
  "id": "23634",
  "username": "noiob",
  "acct": "noiob@awoo.space",
  "display_name": "shork",
```

My proposition is to implement the following change:

- re-map beehive event field `username` from the Mastodon API field `display_name` to `username` (it may introduce regression if a user was wrongly using this field)
- introduce a new beehive event field `display_name` mapped to the Mastodon API field `display_name` (users having a regression will have to use this field instead)
- introduce a new beehive field `acct` mapped to the Mastodon API field `acct` (this field will be used by @Sakrecoer to retrieve the handler)

To be noted, and specified by the Mastodon API documentation, in the case of a local user, the fields `acct` and `username` have the same value.

**Local user**

```
  "id": "1",
  "username": "Gargron",
  "acct": "Gargron",
[...]

PR #367, opened 2021-01-12
MastodonBee - Account mapping fix and enhancement in events

corrantho:
- Fixed account mapping for **reblog** and **mention** events.
   - Used `notif.Account` instead of `notif.Status.Account`
- Extended **follow**, **favourite**, **reblog** and **mention** events with two new fields `display_name` and `acct` respectively mapped to the same mastodon API fields
- breaking modification (to be consistent with Mastodon API) on **follow**, **favourite**, **reblog** and **mention** events: 
    - re-mapped `username` from mastodon  API field`display_name` to mastodon API field `username`.
    
Solve issue: https://github.com/muesli/beehive/issues/356

### DIFF
--- a/bees/mastodonbee/events.go
+++ b/bees/mastodonbee/events.go
@@ -109,7 +109,17 @@ func (mod *MastodonBee) handleNotification(notif *mastodon.Notification) {
 				},
 				{
 					Name:  "username",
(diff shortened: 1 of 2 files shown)


muesli:
Thank you!

